PR #307, opened 2020-12-18
[chore] Ensure accurate error reporting if beforeAll callback fails

bjoerge:
This fixes an issue that reported inaccurate error message if something failed during `beforeAll`: E.g. if launching `qawolf` in the `beforeAll`-callback failed, the `browser` variable would be undefined, which again would make the `afterAll` callback fail when it tried to call `close()` on it. This obscured the original error, and made jest report `TypeError: Cannot read property 'close' of undefined` as the error and not the original error that happened in the beforeAll`-callback.

Note: added a few non-null assertions in the actual tests, since we can be sure that the variables are initialized when if we get to actually running the test.

### DIFF
--- a/.qawolf/responsive.test.ts
+++ b/.qawolf/responsive.test.ts
@@ -1,0 +1,52 @@
+import {Browser, BrowserContext} from 'playwright'
+import qawolf from 'qawolf'
+
+let browser: Browser | undefined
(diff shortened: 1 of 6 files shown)


mariuslundgard (approved):
LGTM

PR #338, opened 2021-01-26
[ui] Prevent Dialog from closing on click outside

mariuslundgard:
### Description

- This change introduces a new `onClickOutside` property to the `Dialog` component.
- This change is introduced because in some cases the click outside behavior must be optional.

### What to review

- Create a Dialog instance without a `onClickOutside` property and make sure the Dialog is not closed when clicking outside of the Dialog.
- Create a Dialog instance​ with a `onClickOutside` property and make sure the callback is called when clicking outside of the Dialog.

### Notes for release

- Adds a new `onClickOutside` property to `Dialog` for handling cases when there's need to trigger changes when clicking outside of the dialog.
**BREAKING CHANGE**: The `onClose` callback is no longer called when the user clicks outside of the `Dialog` component, since there's now a designated callback for that.

### DIFF
--- a/ui/src/components/dialog/dialog.stories.tsx
+++ b/ui/src/components/dialog/dialog.stories.tsx
@@ -27,6 +27,8 @@ export default {
 export const props = () => {
   const header = text('Header', 'Props example', 'Props')
 
+  const onClickOutside = boolean('Close when click outside', false, 'Props')
+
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
Nice, LGTM! ✨

PR #567, opened 2021-03-25
fix(ui): fix `useMediaIndex` browser support issue (Safari 13)

good-idea:
### Description

- Adds support for older mediaQuery syntax
- Fixes https://github.com/sanity-io/design/issues/566

### What to review

I tested this in the local storybook with `dev:ui`. When viewing the `useMediaIndex` entry, it was initially broken, but this change fixed it.

I'm unable to test in my local studio (using `yarn link @sanity/ui`) because of some Styled Components issues.

### Notes for release

- Adds support for Safari 13


mariuslundgard (approved):
Thanks for contributing! ❤️

PR #671, opened 2021-04-15
feat(ui): add role attribute to toast

hermanwikner:
### Description

Adds the role attribute to the toast primitive, in order to enable screen readers to announce the message.

### What to review

Trigger a toast message with screen reader enabled, and the message should be announced.


mariuslundgard (approved):
This is great – thank you!

PR #833, opened 2021-06-10
feat(ui): skeleton components

hermanwikner:


packages/@sanity/ui/src/components/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,48 @@
...
+import styled from 'styled-components'
+import {Box} from '../../primitives'
+import {BoxProps, ResponsiveRadiusProps} from '../../primitives'
+import {responsiveRadiusStyle, ResponsiveRadiusStyleProps} from '../../styles/internal'
+import {skeletonStyle} from './styles'
+
+const Root = styled(Box)<{$animated: boolean; $visible: boolean} & ResponsiveRadiusStyleProps>`
+  ${skeletonStyle}
+  ${responsiveRadiusStyle}
+`
```
mariuslundgard, line 11: I usually use the function syntax when composing style functions:

```tsx
const Root = styled(Box)<{$animated: boolean; $visible: boolean} & ResponsiveRadiusStyleProps>(
  skeletonStyle,
  responsiveRadiusStyle
)
```

Would be nice for consistency, but not a requirement!

packages/@sanity/ui/src/components/skeleton/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const keyframe = keyframes`
+  0% {
+    background-position: 100%;
+  }
+  100% {
+    background-position: -100%;
+  }
+`
+const animation = css`
```
mariuslundgard, line 11: Very nitpick (OCD?): I prefer newlines between variables that take up more space than 1 line. Not a requirement to change this, though.

packages/@sanity/ui/src/components/skeleton/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+  background-image: linear-gradient(
+    270deg,
```
mariuslundgard, line 13: Can we add a comment that says why this value is exactly `270deg` here?

packages/@sanity/ui/src/components/skeleton/textSkeleton.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return <Root {...restProps} $size={$size} ref={ref} $style="text" />
+})
+
+/**
+ * This API might change. DO NOT USE IN PRODUCTION.
+ * @beta
+ */
+export const LabelSkeleton = forwardRef(function TextSkeleton(
+  props: TextSkeletonProps,
```
mariuslundgard, line 53: I think it would be nicer (more discoverable and explicit) if each of the typographic skeleton components have their own props interface:

| Component | Props interface |
|---|---|
| `TextSkeleton` | `TextSkeletonProps` |
| `HeadingSkeleton` | `HeadingSkeletonProps` |
| `CodeSkeleton` | `CodeSkeletonProps` |
| `LabelSkeleton` | `LabelSkeletonProps` |

It would be repeating the same interface, but I think being explicit trumps that. What do you think?

mariuslundgard (commented):
This is really good stuff! I only have minor comments.

mariuslundgard (approved):
Great stuff ✨

PR #871, opened 2021-07-09
feat(ui): add option to hide `Dialog` close button

hermanwikner:
### Description

This update is introduced as we have encountered situations in the studio where we need to hide the close button.

- Adds a `__unstable_hideCloseButton` prop to the `Dialog`.
- Conditionally renders close button depending on the `__unstable_hideCloseButton` value, and whether `onClose` is passed or not.

### What to review

If `onClose` is passed and `__unstable_hideCloseButton` is `true`, the close button should not be rendered.

```
 <Dialog onClose={handleClose} __unstable_hideCloseButton>
    {/* content */}
 </Dialog>
```

If `onClick` is not passed, the close button should not be rendered.

```
 <Dialog>
    {/* content */}
 </Dialog>
```

### Notes for release

- Add `__unstable_hideCloseButton` property to `Dialog` to hide close button

### DIFF
--- a/packages/@sanity/ui/src/components/dialog/dialog.tsx
+++ b/packages/@sanity/ui/src/components/dialog/dialog.tsx
@@ -24,12 +24,15 @@ export interface DialogProps extends ResponsivePaddingProps, ResponsiveWidthProp
    * @beta
    */
   __unstable_autoFocus?: boolean
+  /**
+   * @beta
+   */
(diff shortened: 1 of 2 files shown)


mariuslundgard (approved):
Since you’re on vacation @hermanwikner, I took the liberty of changing the property to a beta feature, and renamed to `__unstable_hideCloseButton`.

PR #971, opened 2021-11-02
fix(MenuItem): add data-disabled to MenuItem when it is disabled

RitaDias:
### Description

Fixes issue where when using the MenuItem as an anchor wasn't giving the right style

### What to review

Changing any of the MenuItem examples to be `as="a"` and making it disabled, before the fix, shows that while it is disabled the styles do not reflect that.

### Notes for release

Fix visual bug when a disabled MenuItem was used as an anchor


mariuslundgard (approved):
LGTM! Thank you ✨

PR #972, opened 2021-11-08
feat(ui): add onClose property to `MenuButton`

hermanwikner:

### DIFF
--- a/packages/@sanity/ui/src/components/menu/__workshop__/onCloseMenuButton.tsx
+++ b/packages/@sanity/ui/src/components/menu/__workshop__/onCloseMenuButton.tsx
@@ -1,0 +1,37 @@
+import {Box, Button, Menu, MenuButton, MenuItem, Stack, useToast} from '@sanity/ui'
+import React, {useCallback} from 'react'
+
+export default function OnCloseMenuButton() {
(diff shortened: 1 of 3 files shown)


packages/@sanity/ui/src/components/menu/__workshop__/onCloseMenuButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+    })
+  }, [push])
+
+  return (
+    <Box padding={[4, 5, 6]}>
+      <Stack space={2}>
+        <MenuButton
+          button={<Button text="With onClose callback" />}
+          id="closable-example"
+          onClose={handleOnClose}
```
mariuslundgard, line 20: Hm, `handleOnClose` feels like double-prefixed naming. Can it be `handleClose`?

mariuslundgard (approved):
Great work – thank you! 🙏

packages/@sanity/ui/src/components/menu/__workshop__/onCloseMenuButton.tsx
hermanwikner: Agree, updated! 👍

PR #973, opened 2022-01-19
refactor(ui): use simpler `isEqual` for `useUnique` hook

rexxars:
### Description

This PR drops the use of `lodash` in `@sanity/ui`, since it was only used for `isEqual` in one place for a very specific use case with the portal implementation, where the comparison was a record of DOM elements - these can be triple-equally checked instead.

This cuts the bundle size for anyone using `@sanity/ui`.

I also took a stab at removing it from the rest of the repo, but stopped when I found `debounce` being used. It's not critical that we replace it in every instance, but it does add a bit of weight. To help tree shaking, I replaced `lodash` with `lodash-es` in the `docs` app.

### What to review

- That the changes makes sense and does not have any outside impact I am unaware of

### Notes for release

- Removed `lodash` as a dependency

### DIFF
--- a/apps/docs/lib/nav/nav.ts
+++ b/apps/docs/lib/nav/nav.ts
@@ -1,4 +1,3 @@
-import {uniq} from 'lodash'
 import {NavItem, NavMenu, NavMenuItem} from './types'
 import {features} from '$config'
 import {isArray, isRecord, isString} from '$lib/types'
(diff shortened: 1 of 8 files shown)


mariuslundgard (approved):
LGTM!

Note that `debounce` is used by the Docs application, not the actual `@sanity/ui` library.

PR #980, opened 2022-05-09
fix(logos): do not pass `dark` prop on to SVG element

rexxars:
### Description

When using `<SanityLogo dark />`, a warning is printed from React because `dark` is not a property on `<svg>`. This is caused by spreading all the props from `SanityLogo` onto `svg`.

This PR fixes this by destructuring the props and removing `dark` before spreading the other props onto the SVG element.

### Notes for release

- Fixed warning when using `dark` prop on `SanityLogo`


mariuslundgard (approved):
Nice – thank you!

PR #985, opened 2022-09-07
refactor: enable react/jsx-runtime and use react 18

stipsan:
### Description

This PR grew a bit in scope, will split up

PR #990, opened 2022-09-07
feat: react `18`

stipsan:
Replaces #985

### Description

Making all things React v18 as baseline.


package.json
```diff
@@ -26,6 +26,9 @@
     "type-check": "tsc --build",
     "watch:packages": "lerna run --stream --parallel --scope @sanity/* watch"
   },
+  "resolutions": {
+    "@types/react": "^18.0.18"
+  },
```
stipsan, line 31: This is due to a known issue in `@types/styled-components`: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/59765

When using `@types/styled-components` together with `@types/react@^18` you need to use yarn `resolutions` or npm v7 `overrides` to dedupe `@types/react`.

mariuslundgard (approved):

mariuslundgard (approved):

package.json
mariuslundgard: Yeah, we do the same in the studio monorepo already

PR #1000, opened 2022-09-13
Add color type to TextInput

PeterDekkers:
### Description

Adds the ability to use the `color` type on a `TextInput` component, which allows for rendering of `<input type="color">` ([docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color)).

### What to review

Even though this is only a small TypeScript typing change, it's worth reviewing whether this type is desirable for the `TextInput` component. I think it is :)

The color input type can be tested in the Workshop under `Primitives > TextInput > Typed`. Alternatively, here is a slightly more elaborate example, using Sanity v3:

```typescript
import React, {ForwardedRef} from 'react'
import {PatchEvent, set} from 'sanity/form'
import {TextInput as _TextInput} from '@sanity/ui'
import {ObjectSchemaType} from '@sanity/types'
import debounce from 'lodash/debounce'
import styled from 'styled-components'

const TextInput = styled(_TextInput)`
    min-height: 2.2em;
`;

interface ColorInputProps {
    onChange: (patchEvent: PatchEvent) => void
    type: ObjectSchemaType
    value?: string
}

const ColorInput = React.forwardRef(
    (props: ColorInputProps, ref: ForwardedRef<HTMLInputElement>) => {
        const {onChange} = props

        const handleChange = React.useCallback(
            debounce(
                (inputValue?: string) => {
                    onChange(PatchEvent.from(set(inputValue)))
                },
                100,
                {leading: true, trailing: true}
            ),
            [onChange]
        )

        return (
            <TextInput
                ref={ref}
                value={props.value || '#000000'}
                type="color"
                clearButton
                onClear={() => handleChange()}
                onChange={(e) => handleChange(e.currentTarget.value)}
            />
        )
    }
)

export default ColorInput
```

This renders the following native, dependency-free color picker in Chrome on macOS, for example:
Screen Shot 2022-09-14 at 11 46 29 AM

### Notes for release

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@typescript-eslint/parser": "^5.37.0",
     "@vitejs/plugin-react": "^2.1.0",
     "cac": "^6.7.14",
-    "chalk": "^5.0.1",
+    "chalk": "^4.1.2",
     "cpx": "^1.5.0",
     "cypress": "^10.8.0",
     "cypress-real-events": "^1.7.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -7079,11 +7079,6 @@ chalk@^4.0.0, chalk@^4.1.0, chalk@^4.1.1, chalk@^4.1.2, chalk@~4.1.0:
     ansi-styles "^4.1.0"
     supports-color "^7.1.0"
 
-chalk@^5.0.1:
-  version "5.0.1"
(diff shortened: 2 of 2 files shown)


PeterDekkers:
Apologies, this PR automatically closed when I pulled in changes from `next`, so I re-opened a new PR at https://github.com/sanity-io/design/pull/1003

PR #1008, opened 2022-09-19
fix: wrong file extension on main field

stipsan:
### Description

This is what's causing the `cannot use import statement outside a module` currently [happening on the react 18 branch](https://github.com/sanity-io/sanity/actions/runs/3056551512/jobs/4930830587).


mariuslundgard (approved):
Yay!

PR #1010, opened 2022-09-27
chore(ci): fix double running ci

stipsan:
### Description

Stops the CI from running everything twice on PRs


mariuslundgard (approved):
Thank you 🙏

PR #1011, opened 2022-09-27
fix(ui): prevent update loops in React v18

mariuslundgard:
### Description

Fixes an issue with state update loops after upgrading to React v18 but refactoring from `useRef` -> `useState`.


stipsan (approved):
Perfect! ✨

PR #1022, opened 2022-12-20
fix: apply display: inline to SVGs to be tailwind compatible

stipsan:
This fix allows removing ugly workarounds like: https://github.com/sanity-io/next-sanity/pull/232

And still have the right SVG alignment:

  
next sanity build_studio_desk

Without this fix, libraries like `next-sanity` has to use ugly CSS workarounds to avoid this outcome on projects that uses Tailwind:
localhost_3000_

### DIFF
--- a/src/primitives/code/styles.ts
+++ b/src/primitives/code/styles.ts
@@ -63,6 +63,12 @@ export function codeBaseStyle(): FlattenInterpolation<StyledThemeProps<Theme>> {
       border-radius: 1px;
     }
 
+    & svg {
+      /* Certain popular CSS libraries changes the defaults for SVG display */
+      /* Make sure SVGs are rendered as inline elements */
+      display: inline;
+    }
+
     & [data-sanity-icon] {
       vertical-align: baseline;
     }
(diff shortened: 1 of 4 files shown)


src/primitives/code/styles.ts
```diff
@@ -64,6 +64,7 @@ export function codeBaseStyle(): FlattenInterpolation<StyledThemeProps<Theme>> {
     }
 
     & [data-sanity-icon] {
+      display: inline;
       vertical-align: baseline;
     }
```
mariuslundgard, line 69: Sanity UI assumes any SVG within `<Code>`, `<Text>`, `<Heading>` and `<Label>` are rendered as inline elements. So I suggest defining display for all SVGs, not only Sanity icons.

```suggestion
    & svg {
      /* Certain popular CSS libraries changes the defaults for SVG display */
      /* Make sure SVGs are rendered as inline elements */
      display: inline;
    }

    & [data-sanity-icon] {
      vertical-align: baseline;
    }
```

mariuslundgard (approved):
Thank you, @stipsan  🙏

PR #1023, opened 2022-12-23
fix: full SSR hydration support

stipsan:
Fixes SSR hydration mismatch errors when using `import {Studio} from 'sanity'` in embeddable scenarios such as Next.js or Remix.


src/components/toast/toastProvider.tsx
```diff
@@ -51,25 +50,29 @@ let toastId = 0
  */
 export function ToastProvider(props: ToastProviderProps): React.ReactElement {
   const {children, padding = 4, paddingX, paddingY, zOffset} = props
-  const [state, setState] = useState<ToastState>({toasts: []})
+  const [state, _setState] = useState<ToastState>([])
+
   const toastsRef = useRef<{[key: string]: {timeoutId: NodeJS.Timeout}}>({})
 
   const push = useCallback((params: ToastParams) => {
+    // Wrap setState in startTransition to allow React to give FormBuilder state updates higher priority
```
mariuslundgard, line 58: Ideally we would not refer to Studio specifics here. I imagine this is useful in a broader range of scenarios too? Can we rephrase?

mariuslundgard (approved):
this is gold 🏅

mariuslundgard (approved):
Good to go!

PR #1050, opened 2023-01-24
feat: add `usePrefersReducedMotion` hook

stipsan:
```tsx
import {usePrefersReducedMotion} from '@sanity/ui'

const reduceMotion = usePrefersReducedMotion()
// true if the OS is setup to reduce motion, false if animations should run
```


mariuslundgard (approved):
Awesome, thanks for this!

PR #1052, opened 2023-01-25
feat: dialog and layer improvements

hermanwikner:


src/utils/layer/layer.tsx
```diff
@@ -24,11 +35,56 @@ const LayerChildren = forwardRef(function LayerChildren(
...
+
+    isTopLayerRef.current = isTopLayer
+  }, [isTopLayer, lastFocusedElement, onActivate])
+
+  const handleFocus = useCallback(
+    (event: FocusEvent<HTMLDivElement, Element>) => {
+      // Call the user-provided onFocus handler if any
+      onFocus?.(event)
+
+      const containsActiveElement = rootElement?.contains(document.activeElement)
```
mariuslundgard, line 63: I’m wondering if there can be cases where the `rootElement` itself is the active element.

E.g. when you use the `Layer` component with a `tabIndex` property: `<Layer tabIndex={0}>`.

_Think_ we should use the new `containsOrEqualsElement` helper for this:

```suggestion
      const containsActiveElement = rootElement && containsOrEqualsElement(rootElement, document.activeElement)
```

src/utils/layer/layer.tsx
```diff
@@ -24,11 +35,56 @@ const LayerChildren = forwardRef(function LayerChildren(
...
+  const {children, onActivate, onFocus, style = EMPTY_RECORD, ...restProps} = props
+  const {zIndex, isTopLayer} = useLayer()
+  const [lastFocusedElement, setLastFocusedElement] = useState<HTMLElement | null>(null)
+  const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null)
+
+  const forwardedRef = useForwardedRef(ref)
+  const isTopLayerRef = useRef<boolean>(isTopLayer)
+
+  // When the layer very first mounts, it will be the top layer, but we don't want to fire the callback in that case.
+  // We use a ref to track the previous value of isTopLayer to determine if the layer has become the top layer since the last render.
```
mariuslundgard, line 47: Nitpick: lines (even comments) should not exceed 100 characters.

src/utils/layer/layer.tsx
```diff
@@ -24,11 +35,56 @@ const LayerChildren = forwardRef(function LayerChildren(
...
+  useEffect(() => {
+    const becameTopLayer = isTopLayerRef.current !== isTopLayer && isTopLayer
+
+    if (becameTopLayer) {
+      onActivate?.({activeElement: lastFocusedElement || null})
```
mariuslundgard, line 52: The `|| null` part is not necessary here:

```suggestion
      onActivate?.({activeElement: lastFocusedElement})
```

src/utils/layer/layer.tsx
hermanwikner: Good point, I agree 👍  I validated this by testing with a `<Layer tabIndex={0} />`. I think it's a rare situation, but it's nice that we support it!

mariuslundgard (approved):
Looks very good. Only a few minor comments.

src/components/dialog/dialog.tsx
```diff
@@ -283,32 +330,67 @@ export const Dialog = forwardRef(function Dialog(
...
+        focusFirstDescendant(cardElement)
 
-      return
-    }
-  }, [])
+        return
+      }
+
+      if (isHTMLElement(event.target)) {
+        focusedElementRef.current = event.target
```
mariuslundgard, line 355: Can we somehow use `activeElement` from layer instead?

mariuslundgard: On second thought, this is fine for now.

PR #1099, opened 2023-05-19
test(ui): add popover position workshop

sjelfull:
We have one or more issues in the Sanity Studio where the position of the reference element of a popover changes due to external factors that doesn't trigger a reposition. Floating UI will listen for viewport resize or scroll events, but have a [`update`](https://floating-ui.com/docs/react#manual-updating) method exposed by the `useFloating` hook for manual updating in those cases where the built-in detection isn't enough.

An example of this issue is when panes is toggled in the Sanity Studio. The actual viewport size didn't change, and there was  no scroll events, so the popover will end up misaligned.

This issue can be replicated in the Popover/Position story inside the workshop in this PR.

This PR adds an optional context for Popover, and exposes a `forceUpdate` method via a new hook called `usePopover` that allows consumers to trigger an update.

It also adds a hook called `useOnForcedUpdate` to be used inside the popover component, that will trigger the update. This should not trigger any errors if you are not using the popover inside a Popover context.

**Questions:**
- Does the `forceUpdate` name make sense, or should we call it something else?
- Should `popover` be moved to `components`?

[sc-32108]

### DIFF
--- a/src/primitives/popover/__workshop__/PositionStory.tsx
+++ b/src/primitives/popover/__workshop__/PositionStory.tsx
@@ -1,10 +1,13 @@
+import {useBoolean} from '@sanity/ui-workshop'
 import {useCallback, useMemo, useState} from 'react'
+import {BoundaryElementProvider} from '../../../utils'
 import {Button} from '../../button'
 import {Card} from '../../card'
-import {Container} from '../../container'
 import {Flex} from '../../flex'
 import {Text} from '../../text'
 import {Popover, PopoverProps} from '../popover'
+import {PopoverProvider} from '../popoverProvider'
+import {usePopover} from '../usePopover'
 
 function PopoverComponent({
   referenceElement,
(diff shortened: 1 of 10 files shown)


hermanwikner (commented):
I suggest that we implement a fix only in the popover component, and without introducing a new React context. By doing so, users won't have to figure out that they need to use a force update function from `usePopover` when using popover with `referenceElement`.

A solution that I think works well is to set up a `ResizeObserver` when `referenceElement` is present. In the callback, we call the `update` function from `useFloating`. With this approach there is no need for a new React context and no additional logic is required to reposition the popover when used with `referenceElement`.

In `popover.tsx`:
```ts
export function Popover() {
  // ...

  const {..., update} = useFloating()

  useEffect(() => {
    if (!referenceElement || !boundaryElement) return

    const ro = new ResizeObserver(update)
    ro.observe(boundaryElement)

    return () => {
      ro.unobserve(boundaryElement)
      ro.disconnect()
    }
  }, [boundaryElement, referenceElement, update])

  return (
    // ...
  )
}
```

sjelfull:
> I suggest that we implement a fix only in the popover component, and without introducing a new React context. By doing so, users won't have to figure out that they need to use a force update function from `usePopover` when using popover with `referenceElement`.
> 
> A solution that I think works well is to set up a `ResizeObserver` when `referenceElement` is present. In the callback, we call the `update` function from `useFloating`. With this approach there is no need for a new React context and no additional logic is required to reposition the popover when used with `referenceElement`.
> 

@hermanwikner The `autoUpdate` middleware already [does this](https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/autoUpdate.ts#L80). The problem in this case is that the reference element, in this case the PTE, doesn’t change its size. There will be less or more whitespace around the PTE, but the size doesn’t change.

PR #1103, opened 2023-06-26
feat: support passing custom component to tree item link

binoy14:
Supports passing custom link component to TreeItem, this is useful when using in a Next.js application where instead of `a` tag you can pass `Link` component from Next. 

Let me know if the naming and the API makes sense.

### DIFF
--- a/src/components/tree/treeItem.tsx
+++ b/src/components/tree/treeItem.tsx
@@ -1,7 +1,7 @@
 import {ToggleArrowRightIcon} from '@sanity/icons'
 import {createElement, memo, useCallback, useEffect, useId, useMemo, useRef} from 'react'
 import styled from 'styled-components'
-import {Box, Flex, Text} from '../../primitives'
+import {Box, BoxProps, Flex, Text} from '../../primitives'
 import {ThemeFontWeightKey} from '../../theme'
 import {
   treeItemRootStyle,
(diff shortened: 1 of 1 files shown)


src/components/tree/treeItem.tsx
```diff
@@ -24,6 +24,10 @@ export interface TreeItemProps {
   space?: number | number[]
   text?: React.ReactNode
   weight?: ThemeFontWeightKey
+  /**
+   * Allows passing a custom element type to the link component
+   */
+  linkAs?: BoxProps['as']
```
mariuslundgard, line 30: First off – this makes a lot of sense – good addition 👍 

Nitpick: all the properties in `@sanity/ui` are alphabetically sorted, since it aids scannability/readability. So would love it if this was place between the `icon` and `padding` property 🙏

src/components/tree/treeItem.tsx
```diff
@@ -157,6 +162,7 @@ export const TreeItem = memo(function TreeItem(
           ref={treeitemRef}
           role="treeitem"
           tabIndex={tabIndex}
+          as={linkAs}
```
mariuslundgard, line 165: This property should also be sorted 🙏

mariuslundgard (approved):

RitaDias (approved):
✅

PR #1104, opened 2023-06-30
fix: issue with tabbing on tree

RitaDias:
👋 
There was an issue where when focusing on a tree element when coming from an outside element (such an input), the ref in the tree wasn't being updated (which in turn broke the navigation on the tree).

You can see the current issue [here](https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJztkM0KwjAQhM%2FmKQbPQrHnNKA3bx58gWhWLWgS0i0tlL67tf4U6w960JO5ZNl8M9kdOXUlvDYmtZukims1D5Rl4C0htT5nrNOQ8ejYsNDMtPcMdmC9RAQdgitgXGHhbCviQCSjxlQJISc5u5Xb%2Bx0xIVJCLppXJYC2mDVmoNJra8gkFYecajCVnFRDvVwN6yP5Bju%2BkMD56iSdXdxRkNEFUD9UvGny8a9XQR8XGNy2RM%2FjSaDxP9Cn%2Bd2H92jc17v0p%2Fgm3dvhVKtBew7YFyaR&title=TreeItem)

### DIFF
--- a/src/components/tree/__workshop__/tabFromElement.tsx
+++ b/src/components/tree/__workshop__/tabFromElement.tsx
@@ -28,7 +28,7 @@ export default function BasicStory() {
       </Box>
       <Wrapper>
         <TextInput />
-        <Tree ref={ref} space={1}>
+        <Tree ref={ref} space={1} onFocus={() => console.log("I'm focus!")}>
           <TreeItem onClick={handleClick} expanded text="Fruit">
             <TreeItem
               data-testid="oranges"
--- a/src/components/tree/__workshop__/tabFromElement.tsx
+++ b/src/components/tree/__workshop__/tabFromElement.tsx
@@ -17,6 +17,10 @@ export default function BasicStory() {
     if (testid) setId(testid)
   }, [])
 
+  const handleFocus = useCallback((event: React.FocusEvent<HTMLDivElement>) => {
+    console.log("I'm focused")
+  }, [])
+
   return (
     <Box padding={[4, 5, 6]}>
       <Box paddingY={3}>
@@ -28,7 +32,7 @@ export default function BasicStory() {
       </Box>
       <Wrapper>
         <TextInput />
-        <Tree ref={ref} space={1} onFocus={() => console.log("I'm focus!")}>
+        <Tree ref={ref} space={1} onFocus={handleFocus}>
           <TreeItem onClick={handleClick} expanded text="Fruit">
             <TreeItem
               data-testid="oranges"
--- a/src/components/tree/__workshop__/tabFromElement.tsx
+++ b/src/components/tree/__workshop__/tabFromElement.tsx
@@ -1,5 +1,5 @@
 import {LinkIcon} from '@sanity/icons'
-import {Box, TextInput, Tree, TreeItem} from '@sanity/ui'
+import {Box, TextInput, Tree, TreeItem, Text} from '@sanity/ui'
 import {usePerfTest} from '@sanity/ui-workshop/plugin-perf'
 import {useCallback, useState} from 'react'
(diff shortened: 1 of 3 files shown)


binoy14 (commented):
works for me but I am not too familiar with this code so will defer to @mariuslundgard for ✅

src/components/tree/__workshop__/tabFromElement.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    if (testid) setId(testid)
+  }, [])
+
+  return (
+    <Box padding={[4, 5, 6]}>
+      <Box padding={2}>
+        This example is to demonstrate that when you tab from an outside element (using the keyboard
+        to navigate), you can still access the tree and tree item. Press the input beneath and start
+        tabbing / using the arrow.
```
mariuslundgard, line 25: Would be good if you could wrap this in a `<Text>` component so it looks better :)

src/components/tree/tree.tsx
```diff
@@ -183,6 +183,10 @@ export const Tree = memo(
       [itemElements]
     )
 
+    const handleOnFocus = useCallback((event: React.FocusEvent<HTMLDivElement>) => {
+      setFocusedElement(event.target as HTMLElement)
```
mariuslundgard, line 187: Good catch to find this bug!

It would be somewhat better to use React's `.currentTarget` property so we can avoid type casting.

And we also need to make sure the `onFocus` property which can potentially be passed to `Tree` is called.

```suggestion
      setFocusedElement(event.currentTarget)
      
      // Call the element's `focus` handler
      onFocus?.(event)
```

src/components/tree/tree.tsx
```diff
@@ -202,6 +206,7 @@ export const Tree = memo(
           ref={forwardedRef}
           role="tree"
           space={space}
+          onFocus={handleOnFocus}
```
mariuslundgard, line 209: Nitpick: The properties here should be a alphabetized, so please move it above the `onKeyDown` property 🙏

src/components/tree/tree.tsx
```diff
@@ -183,6 +183,10 @@ export const Tree = memo(
       [itemElements]
     )
 
+    const handleOnFocus = useCallback((event: React.FocusEvent<HTMLDivElement>) => {
```
mariuslundgard, line 186: Nitpick: rename to `handleFocus` 🙏

src/components/tree/tree.tsx
RitaDias: Hmm replacing with currentTarget will point me towards the TreeItem and not the Tree component which will then break the behavior again :/
That being said, when adding the rest I noticed that I don't need to cast 👍

mariuslundgard (approved):
Tiny issue in a comment. Otherwise looks good 👍

src/components/tree/tree.tsx
```diff
@@ -183,6 +183,16 @@ export const Tree = memo(
...
     )
 
+    const handleFocus = useCallback(
+      (event: React.FocusEvent<HTMLDivElement>) => {
+        setFocusedElement(event.target)
+
+        // Call the element's `focus` handler
+        props.onFocus?.(event)
+      },
+      [props]
```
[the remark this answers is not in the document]
RitaDias: huh interesting 🤔 that was the first thing that I wrote as well but then I got a warning from vscode (image)
image

I'll change it and hopefully nothing on the CI breaks 😂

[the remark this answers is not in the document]
mariuslundgard: You may also destructure it from the props

PR #1106, opened 2023-07-06
feat: menu improvements

hermanwikner:
This pull request fixes an issue where the menu items were incorrectly focused when hovering between them while using `shouldFocus` with the `Menu` component. See video below of the issue.

https://github.com/sanity-io/ui/assets/15094168/d82364e0-a74a-44f2-aef2-e09542f036e7

### DIFF
--- a/src/components/menu/useMenuController.ts
+++ b/src/components/menu/useMenuController.ts
@@ -173,9 +173,14 @@ export function useMenuController(props: {
   )
 
   const handleItemMouseLeave = useCallback(() => {
+    // Set the active index to -2 to deactivate all menu items
+    // when the user moves the mouse away from the menu item.
+    // We avoid using -1 because it would focus the first menu item,
+    // which would be incorrect when the user hovers over a gap
+    // between two menu items or a menu divider.
+    setActiveIndex(-2)
     rootElement?.focus()
-    setActiveIndex(-1)
-  }, [rootElement, setActiveIndex])
+  }, [setActiveIndex, rootElement])
 
   // Set focus on the currently active element
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/components/menu/menuItem.tsx
```diff
@@ -21,6 +21,7 @@ import {useMenu} from './useMenu'
  */
 export interface MenuItemProps extends ResponsivePaddingProps, ResponsiveRadiusProps {
   as?: React.ElementType | keyof JSX.IntrinsicElements
+  closeMenuOnClick?: boolean
```
mariuslundgard, line 24: Is this one still needed?

mariuslundgard:
@hermanwikner 

To clarify – in what use case do we not want to close the menu when a selection is made?

src/components/menu/useMenuController.ts
```diff
@@ -173,9 +173,13 @@ export function useMenuController(props: {
   )
 
   const handleItemMouseLeave = useCallback(() => {
-    rootElement?.focus()
```
mariuslundgard, line 176: Is removing this required to fix the bug?

hermanwikner: It seems to work as expected with and without it. We can keep it – but I can't see why it is needed? Do you know in what situations it is needed?

mariuslundgard: Yes, if you move the mouse in between 2 items, it will look like no item is currently focused. So if you press ENTER in that situation, you will end up selecting something without intending to.

hermanwikner: Good point, added it back again

mariuslundgard (approved):
💯

PR #1107, opened 2023-07-07
fix: allow for enter to close / open items in treeItem

RitaDias:
👋 
There was an issue where when using the keyboard to navigate through a item list you weren't able to close the item tree. I'm unsure if this is something that should be currently allowed but it made sense especially while we were trying to transverse large multiple item lists (that they themselves have tree items that can close / open)

You can see the current issue [here](https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJztkM0KwjAQhM%2FmKQbPQrHnNKA3bx58gWhWLWgS0i0tlL67tf4U6w960JO5ZNl8M9kdOXUlvDYmtZukims1D5Rl4C0htT5nrNOQ8ejYsNDMtPcMdmC9RAQdgitgXGHhbCviQCSjxlQJISc5u5Xb%2Bx0xIVJCLppXJYC2mDVmoNJra8gkFYecajCVnFRDvVwN6yP5Bju%2BkMD56iSdXdxRkNEFUD9UvGny8a9XQR8XGNy2RM%2FjSaDxP9Cn%2Bd2H92jc17v0p%2Fgm3dvhVKtBew7YFyaR&title=TreeItem) 

https://github.com/sanity-io/ui/assets/6951139/55c3cd60-ae2f-4c0d-892a-659e74852bda

This will fix it (follows the same actions as the "onClick"

### DIFF
--- a/src/components/tree/treeItem.tsx
+++ b/src/components/tree/treeItem.tsx
@@ -89,7 +89,7 @@ export const TreeItem = memo(function TreeItem(
 
       if (
         target instanceof HTMLElement &&
-        (target.getAttribute('data-ui') === 'TreeItem__box' ||
+        (target.getAttribute('data-ui') === 'TreeItem' ||
           target.closest('[data-ui="TreeItem__box"]'))
       ) {
         event.stopPropagation()


src/components/tree/treeItem.tsx
```diff
@@ -105,10 +105,13 @@ export const TreeItem = memo(function TreeItem(
       if (focused && event.key === 'Enter') {
         const el = treeitemRef.current || rootRef.current
 
+        event.stopPropagation()
+        setExpanded(itemKey, !expanded)
+        setFocusedElement(rootRef.current)
         el?.click()
```
binoy14, line 111: ❓I am not very familiar with this code but doing an `el.click()` would trigger the handleClick? But I am guessing that it does not match the if conditional there so it doesn't change the state. Is the click still needed if this function is doing the same state updates?

RitaDias: Good point/question (specifically about if the click is needed)! 
I tried removing it and everything seems to work as expected but at the same time since this is the only piece of code that was already existing in there, I'm not sure if it's serving another purpose that we're just not seeing 🤔 

I am thinking, maybe it's a better alternative:
On the condition, specifically line [92](https://github.com/sanity-io/ui/blob/feature/sc-34577/enter/tree/src/components/tree/treeItem.tsx#L92), the condition looks to me like it should be `TreeItem` (which would fix it) instead of 'TreeItem__box'.

Making this change would get the code working _and_ would be cleaner 🤔 what do you think?

binoy14: Makes senes to me, if this causes issues then it's not too hard to fix it later

RitaDias: Thanks! Done :)

binoy14 (approved):
🚢

PR #1108, opened 2023-07-07
fix(primitives): close tooltip on button click

ninaandal:
If a tooltip was added to a button that opened a popover for instance, the tooltip would show underneath the popover. This pr adds a a `closeOnClick` prop that closes the tooltip when the child element is clicked. 

Clipboard 2023-26-06 at 11 08 18 AM

### DIFF
--- a/src/primitives/tooltip/__workshop__/props.tsx
+++ b/src/primitives/tooltip/__workshop__/props.tsx
@@ -12,6 +12,7 @@ export default function PropsStory() {
   const placement = useSelect('Placement', WORKSHOP_PLACEMENT_OPTIONS, 'top')
   const portal = useBoolean('Portal', true)
   const shadow = useSelect('Shadow', WORKSHOP_SHADOW_OPTIONS, 2)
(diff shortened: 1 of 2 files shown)


mariuslundgard (approved):
Thanks, this works!

Although, I think an even more complete solution would be to enable/disable the tooltip depending on whether the popover menu is open or not. It can be achieved with the `onOpen` and `onClose` properties on `<MenuButton>`:

```tsx
import {AddIcon} from '@sanity/icons'
import {Button, Menu, MenuButton, MenuItem, Text} from '@sanity/ui'

function Navbar() {
  const [menuOpen, setMenuOpen] = useState(false)

  return (
    <MenuButton
      button={
        <Tooltip content={<Text size={1}>Create</Text>} disabled={menuOpen} padding={2}>
          <Button aria-label="Create" icon={AddIcon} mode="bleed" />
        </Tooltip>
      }
      menu={<Menu>...</Menu>}
      onClose={() => setMenuOpen(false)}
      onOpen={() => setMenuOpen(true)}
    />
  )
}
```

PR #1111, opened 2023-07-24
fix(components): add maxwidth to tabs

ninaandal:
Tabs had horizontal scroll when the title is very long, there was no limit to the width. This pr customise the `Inline` component for the `tabList` to prevent the tabs to overflow the width of the parent container. 

Screenshot 2023-07-24 at 15 21 36

Note: should the children of the inline object always have a maxWidth of 100%? This would make sense in most cases, but not sure if it should be added to every case..


mariuslundgard (approved):
Thanks – this solves the problem! As you mention, this could be applied directly to the `Inline` component.

PR #1119, opened 2023-08-14
fix(tooltip): use floatingStyles to prevent rerenders of tooltip

ninaandal:
The tooltip would rerender unnecessarily when hovering over it. 

The `x` positioning used in `rootstyles` was most likely the reason for this, as the `x` position would be recalculated several times, causing the tooltip placement to "jump". 
By using the `floatingStyles` from `floating-ui', this was fixed. The size of the tooltip is calculated using the `size` from `floating-ui`.

### DIFF
--- a/src/primitives/tooltip/__workshop__/props.tsx
+++ b/src/primitives/tooltip/__workshop__/props.tsx
@@ -7,10 +7,7 @@ import {
 } from '../../../__workshop__/constants'
 
 export default function PropsStory() {
-  const text =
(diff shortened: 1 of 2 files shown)


stipsan (approved):
Not an expert on floating ui but this looks good to me 👍

mariuslundgard (commented):
I looked at the preview deployment, and it looks like tooltips (still?) get cut off.
https://sanity-ui-workshop-git-fix-tooltip-hover.sanity.build/primitives/tooltip/props?scheme=light&viewport=320&value=eJyrVnLOzytJzStRslIKyc/PKcksUEiGiCgkFuekZCfmKuRkpyTmFiuk5AA5xUASKJGbkp2jAOEBSaB8MQgp1QIAkZsdxA==

Wondering if we need to subtract the boundary padding to get it exactly right?

image

Loving the increase in no. of PRs to `@sanity/ui`! ❤️

mariuslundgard (approved):

PR #1130, opened 2023-08-31
fix(tooltip): add default fallbackplacements

ninaandal:
Both the `Tooltip` and the `Popover` had some issues with the fallback placements. 

This hook: `const fallbackPlacements = useArrayProp(fallbackPlacementsProp)` was returning an empty array, and as the default was supposed to be the oppside of this, it didn't apply a default value. 

This resulted in various issues in the studio where a fallback placement was not set when the `popover` or `tooltip` didn't fit the element. 

Note: it might be unnecessary to add all possible alternatives in the fallback placement array, so let me know if I should remove some


src/primitives/popover/popover.tsx
```diff
@@ -89,7 +89,20 @@ export const Popover = memo(
...
+        'bottom',
+        'bottom-start',
+        'bottom-end',
+        'left',
+        'left-start',
+        'left-end',
+        'right',
+        'right-start',
+        'right-end',
+      ],
```
mariuslundgard, line 105: I'd move this value to a `DEFAULT_FALLBACK_PLACEMENTS` variable outside of the component scope and do
```suggestion
      fallbackPlacements = DEFAULT_FALLBACK_PLACEMENTS,
```

src/primitives/tooltip/tooltip.tsx
```diff
@@ -65,7 +65,20 @@ export const Tooltip = forwardRef(function Tooltip(
...
+      'bottom',
+      'bottom-start',
+      'bottom-end',
+      'left',
+      'left-start',
+      'left-end',
+      'right',
+      'right-start',
+      'right-end',
+    ],
```
mariuslundgard, line 81: same as above

mariuslundgard (approved):
This works and improves things 👍 

Ideally we would have a map between placements and their default fallback placements. E.g. 

```ts
const DEFAULT_FALLBACK_PLACEMENTS: Record<Placement, Placement[]> = {
  top: ['bottom', 'left', 'right'],
  'top-start': ['bottom-start', 'left-start', 'right-start'],
  // etc.
}
```

And use that in the popover + tooltip components as (pseudo code):

```ts
fallbackPlacements = props.fallbackPlacements ?? DEFAULT_FALLBACK_PLACEMENTS[props.placement ?? 'top']
```

src/primitives/popover/popover.tsx
```diff
@@ -37,6 +37,21 @@ import {PopoverUpdateCallback, PopoverWidth} from './types'
 
 export type {PopoverUpdateCallback}
 
+const DEFAULT_FALLBACK_PLACEMENTS: Record<Placement, Placement[]> = {
```
pedrobonamin, line 40: Great addition! 
Probably we should move it to another file and share it between the popover and the tooltip, it could be a new file `src/primitives/constants.ts` 
What do you think?

ninaandal: The `Popover` already had a `constants.ts` file, so added it there and created a new one for the `Tooltip` ☺️

PR #1134, opened 2023-09-08
feat(tooltip): allows for the tooltip to be dismissed on Escape

pedrobonamin:
Allows for the tooltip to be dismissed on Escape.
Requirement for WCAG 1.4.13 
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

### DIFF
--- a/src/primitives/tooltip/tooltip.tsx
+++ b/src/primitives/tooltip/tooltip.tsx
@@ -177,23 +177,27 @@ export const Tooltip = forwardRef(function Tooltip(
   const closeDelay = isInsideGroup ? delayGroupContext.closeDelay : closeDelayProp
 
   const handleIsOpenChange = useCallback(
-    (open: boolean) => {
+    (open: boolean, immediate?: boolean) => {
       if (isInsideGroup) {
         //  When it's inside a group, the open or close status will be handled by the group.
         if (open) {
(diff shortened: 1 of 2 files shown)


src/primitives/tooltip/tooltip.tsx
```diff
@@ -243,6 +243,22 @@ export const Tooltip = forwardRef(function Tooltip(
   // Update reference
   useEffect(() => refs.setReference(referenceElement), [referenceElement, refs])
 
+  useEffect(() => {
+    // If the user clicks on escape key, close the tooltip.
+    if (!showTooltip) return
+
+    function handleWindowKeyDown(event: KeyboardEvent) {
+      if (event.key === 'Escape') {
+        handleIsOpenChange(false)
```
robinpyon, line 252: Tooltips dismissed with ESC should close immediately and disregard any existing `delay` value – would you agree?

If so, perhaps `handleIsOpenChange` could potentially accept an optional `immediate` param (or similar) which if set, forces all delay values within to be set to `0` and limit this behaviour to ESC keypresses only.

robinpyon (approved):
Thanks @pedrobonamin!

mariuslundgard (approved):

PR #1150, opened 2023-10-01
feat: add radius='full' support and supporting stories

robinpyon:
### Description
This PR adds support for `radius="full"` on primitives and components that support it. Full radius in this case is `border-radius: 9999px` rather than a percentage, since primitives may be non-square.

**Note:** this support has carried over to `popoverRadius` in `<MenuButton>` components too. Setting a larger radius on these don't render particularly well (you can view this for yourself in the `MenuButton/PopoverRadius` story). This isn't a regression ([example](https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJyFUc1qAyEQvu9TDF6SUMKSpvQQdKFNLj2UQt%2FArsNGcFXc2ZAivnvVpA300N7Gz%2B9vlO9lUOClUtoOIj4kmOjToIiR8ExPRg92B4seLWFYpNQ1%2FBXt%2FDwTOdsAfNRBRH5BoIgEu1FgaR3pHsGdMBjpIUPOk3Z2gg3cwXbFoO1SdvLOF867VHqeRHwsmFaCjdlrfYlZ41mO3iDLVwUWcZkngFqpq%2BP18EI4Xru81TTYlJy%2FKfe%2FKQd90grDv8Ltj5C331VWdScje8xNMzXo4Ujstmd%2BYO8CSbMDCjOm1GQL3pbf6L4ACAJ%2FoA%3D%3D&title=MenuButton+popoverRadius+example)), but this update can make this exceptionally noticeable when using `popoverRadius="full"`. I think it's OK to leave this behaviour as-is – for instance, users could always opt-in to use `full` radii in combination with a popover with generous padding.

**Storybook additions:**
- Basic component stories for `<Autocomplete>`, `<Hotkeys>` and `<MenuButton>`
- Dedicated `radius` prop stories for applicable primitives and components
- Story controls are now generated using values defined in `studioTheme` – leading to less hard-coding across the board
- Similarly, added a `radiusBuilder` helper to generate Radius stories using values defined in the current `studioTheme`


stories/primitives/Popover.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  render: (props) => {
+    return (
+      <Card paddingY={6}>
+        <Popover {...props} />
+      </Card>
+    )
+  },
+}
+
+// @todo: understand why this story doesn't render in storybook docs
```
pedrobonamin, line 34: I do see it in storybook docs, maybe needed a restart on the server?

pedrobonamin (approved):

PR #1151, opened 2023-10-02
chore(storybook): remove unsightly story/docs background flashes, add favicon

robinpyon:
### Description

This PR hides white BG flashes on story / docs pages. It also loads our current web favicon – perhaps we may want to add a Storybook / UI specific one in future?

**Separate to this PR, but tangentially linked:** a separate Vercel project (`sanity-ui-storybook`) has been created which will build Storybook on commits to `feat/facelift-mvi-2` only. This is accomplished with a configured **Ignored Build Step** in the project and done to prevent noise on other pre-facelift-2 branches.

Build + install commands for `sanity-ui-storybook` are defined Vercel-side (rather than in `vercel.json`) to allow `ui-workshop` to continue to be built as normal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist
 node_modules
 
 /etc
+storybook-static
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -1,3 +1,10 @@
+<!-- prettier-ignore -->
+<link rel="icon" href="https://www.sanity.io/static/images/favicons/favicon-96x96.png" sizes="96x96" type="image/png" />
+<!-- prettier-ignore -->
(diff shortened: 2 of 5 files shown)


pedrobonamin (approved):
Thanks for this @robinpyon 
Question: why you don't want to build the rest of facelift branches? I think it gives us the added value to review the branches faster if they are also deployed.

robinpyon:
@pedrobonamin Just clarifying what we discussed separately, we're limiting these deployments to `feat/facelift-mvi-2` only since we don't have failed runs on non-facelift branches. But given the limited activity on this repo, it might be OK to disable this – we'll play it by ear!

PR #1152, opened 2023-10-04
chore(cssvars): replace color theming through styled-components for css vars.

pedrobonamin:
## To discuss:

- Is there any other components / stories worth considering?
- All suggestions are appreciated and welcome !

## Worth considering:
- This PR introduces css variables for theming, removing the need for `styled-components` the `<StyledThemeProvider>` context could be removed with further changes.
- Removes all theme colors related files.
- Components that don't open open the possibility of defining tones, should always use the css variables in the `default` value, this ones are going to be mutated by the parents.
  e.g. the shadow doesn't receive a tone, so it's always inherited from the parents, in this cases we want to always use the default variable, as these ones will be overridden by a parent element if necessary

 ```
 function shadowStyle(shadow: ThemeShadow | null): CSSObject {
  const outline = `0 0 0 ${rem(1)} ${cssVars.default['base-shadow-outline-color']}`
  const umbra = toBoxShadow(shadow.umbra, cssVars.default['base-shadow-umbra-color'])
  const penumbra = toBoxShadow(shadow.penumbra, cssVars.default['base-shadow-penumbra-color'])
  const ambient = toBoxShadow(shadow.ambient, cssVars.default['base-shadow-ambient-color'])

  return {boxShadow: `${outline}, ${umbra}, ${penumbra}, ${ambient}`}
}

```
 e.g if we have a element that supports tone like the badge, it will use the variable according to the tone received, and not the default one. If not tone is passed it will fallback to the default, so it will also inherit the variable from the parent as in the previous scenario.
```
export function badgeStyle(props: BadgeStyleProps): CSSObject {
  const {$mode, $tone} = props. 

  return {
    // backgroundColor: color.enabled.bg,
    backgroundColor: $mode === 'outline' ? cssVars[$tone]['bg-base'] : cssVars[$tone]['bg-tint'],
    color: cssVars[$tone]['text-secondary'],            
    boxShadow: $mode === 'outline' ? `inset 0 0 0 1px ${cssVars[$tone]['border-base']}` : undefined,
    cursor: 'default',

    '&:not([hidden])': {
      display: 'inline-block',
    },
  }
}

```
- Cards won't insert a new themeProvider, color theming will be handled by css vars, when adding a card with a defined tone or scheme, this card will create a new internal context of css vars.
  - The default vars will be generate using the tone provided. So, if we add a tone of caution, the var `--default-text-primary` will use the value of `--caution-text-primary`.
  - When a card defines a tone, all variables that require mixing will be mixed with the new tone default background.
- [mutable card variables](https://github.com/sanity-io/ui/pull/1152/files#diff-8dc3c073a170919c25f5069a2ac0e52f9f117dbdb5474868ad916df59fc84213R10) are divided because this variables are updated in multiple places, not only the card component.
  - A use case of this type of variables is the fg color, which is updated by the button component and affects the text elements that are inside that button component.
- The file [colorVarsStyle](https://github.com/sanity-io/ui/pull/1152/files#diff-8dc3c073a170919c25f5069a2ac0e52f9f117dbdb5474868ad916df59fc84213) doesn't remove legacy variables for backwards compatibility 
- css variables shouldn't be hardcoded in the css files, all of them should be used from one of the available objects, this means there shouldn't be references to `var(xxxxx)`in the css files, unless the variable is defined in that same file. This will allow us to move faster when refactoring variables and to be type safe.

- Tooltips need to define a new css vars context, as they are mounted into a portal in an html element that it's outside of the scope of the current card that contains the updated css vars, that's why `toneContext` is introduced. See [`<popoverCard>`](https://github.com/sanity-io/ui/pull/1152/files#diff-5c9878b951934a2cadbd8c826662ed230c95ddcd3a409dc6996b77821f36e6dd)

Introduces global css variables for colors, following the colors specified in the facelift redesign, the same colors shown in `studio/colors` story.

The colors will be added to the body element and will change according to the theme (dark | light) selected. 
Cards introduce the possibility to change the theme , so a new css vars context is added into the cards. 

### Current implementation
Arcade playground with cards with buttons that inherit the tone.
[Sanity arcade](https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJztlcFunDAQhu99CosLl0ZW0p4qzCEbVa2Uqislt6oHL56wVgxGxmw2Qrx7bcCGJd2GS6IKcfF6Zn7PeMZ82uirgCNKaUHqqwYVlDGep%2FeyNeMPCEUbqpjzk%2FpzgxRlvCpJfdmgck%2BZfLJbKzXiO02TR1QWNAGXoPV%2FA2rPx%2F1vhJ3DCe7hqGO7RLjdOv8t3YGI2zXCneFC9uLOMOY1ZSnE7RrhznBKPJZG15XWMkfalCHBDTzQSugA4VfimWRAgp0AYHPF6V6WbeY6RGHjRyH1IzyXyC6k%2FhVutBLhRxTe7fmDtptt%2BLsZBm4eBU%2BmRHclmozJ%2BG55qV9OvoshYSdHgo3MNeTmTiUISDQwUmtVwajEqX6r4MDhadww7kudNvUD8qobhc%2Bza01Sn47oZwE5yoza5my8mDMSWO9Fd%2BoCjjQrBAReYIOk9mZfMh45etd3DZmvpbkpfBmM2%2FuH8Orvwht%2B4AzUzCSfJkkifHrRoedCmLcyfZmziqd7PTRbyEIeQJG6LqTSVHxB7Su5o33%2BCLe0tZBiS%2BmAqxmh%2Bf4KxTOqnoPl0rt1HZ4B0sfn0DsRe3pXcFdw0fuCK0uu%2BQGWTK5v8RyNXjCL3Yl6sfCu3P7H3Ca0sldaMLYb1%2BEZDn18DrQT8cDsCu0Krdu9PbTK%2FHMkVCyZWt%2FiORK9YBa3E%2FUK7gqus98UXPeN%2FwE6LT4G&title=Tones+stories)
Screenshot 2023-10-05 at 18 31 17

### With css variables 
https://sanity-ui-storybook-git-edx-471-use-vars.sanity.build/?path=/docs/tests-tones--docs

Screenshot 2023-10-11 at 16 24 58

### CODE REVIEW

File: src/primitives/card/styles.ts [hunk 1]
```diff
@@ -52,6 +52,7 @@ export function cardColorStyle(props: CardStyleProps & ThemeProps): ReturnType<t
   const border = {width: 0, color: 'var(--card-border-color)'}
 
   return css`
+    ${_cssVarStyles(theme.scheme, theme.tones)}
```
pedrobonamin, line 55: Adds a css vars context to allow changes between light and dark mode inside the cards.

src/theme/lib/theme/color/cssVars.ts
```diff
@@ -0,0 +1,57 @@
...
+const themeNames: ThemeColorName[] = [
+  'default',
+  'transparent',
+  'primary',
+  'positive',
+  'caution',
+  'critical',
+]
+
+export const cssVars: Record<ThemeColorName, Record<ColorKey, string>> = themeNames.reduce(
```
pedrobonamin, line 43: This creates a object with all the available css vars, and allows us to use it providing type safety
 https://epicreact.dev/css-variables/

```
{
    "default": {
        "text_primary": "var(--default-text_primary)",
        "text_secondary": "var(--default-text_secondary)",
        "text_tertiary": "var(--default-text_tertiary)",
        "bg_base": "var(--default-bg_base)",
        "bg_base_hover": "var(--default-bg_base_hover)",
        "bg_base_active": "var(--default-bg_base_active)",
        "bg_accent": "var(--default-bg_accent)",
        "bg_accent_hover": "var(--default-bg_accent_hover)",
        "bg_accent_active": "var(--default-bg_accent_active)",
        "bg_tint": "var(--default-bg_tint)",
        "icon_default": "var(--default-icon_default)",
        "icon_inverted": "var(--default-icon_inverted)",
        "border_base": "var(--default-border_base)",
        "border_base_hover": "var(--default-border_base_hover)",
        "border_accent": "var(--default-border_accent)",
        "border_accent_inverted": "var(--default-border_accent_inverted)"
    },
    "transparent": {
        "text_primary": "var(--transparent-text_primary)",
        "text_secondary": "var(--transparent-text_secondary)",
        "text_tertiary": "var(--transparent-text_tertiary)",
        "bg_base": "var(--transparent-bg_base)",
        "bg_base_hover": "var(--transparent-bg_base_hover)",
        "bg_base_active": "var(--transparent-bg_base_active)",
        "bg_accent": "var(--transparent-bg_accent)",
        "bg_accent_hover": "var(--transparent-bg_accent_hover)",
        "bg_accent_active": "var(--transparent-bg_accent_active)",
        "bg_tint": "var(--transparent-bg_tint)",
        "icon_default": "var(--transparent-icon_default)",
        "icon_inverted": "var(--transparent-icon_inverted)",
        "border_base": "var(--transparent-border_base)",
        "border_base_hover": "var(--transparent-border_base_hover)",
        "border_accent": "var(--transparent-border_accent)",
        "border_accent_inverted": "var(--transparent-border_accent_inverted)"
    },
    "primary": {
        "text_primary": "var(--primary-text_primary)",
        "text_secondary": "var(--primary-text_secondary)",
        "text_tertiary": "var(--primary-text_tertiary)",
        "bg_base": "var(--primary-bg_base)",
        "bg_base_hover": "var(--primary-bg_base_hover)",
        "bg_base_active": "var(--primary-bg_base_active)",
        "bg_accent": "var(--primary-bg_accent)",
        "bg_accent_hover": "var(--primary-bg_accent_hover)",
        "bg_accent_active": "var(--primary-bg_accent_active)",
        "bg_tint": "var(--primary-bg_tint)",
        "icon_default": "var(--primary-icon_default)",
        "icon_inverted": "var(--primary-icon_inverted)",
        "border_base": "var(--primary-border_base)",
        "border_base_hover": "var(--primary-border_base_hover)",
        "border_accent": "var(--primary-border_accent)",
        "border_accent_inverted": "var(--primary-border_accent_inverted)"
    },
    "positive": {
        "text_primary": "var(--positive-text_primary)",
        "text_secondary": "var(--positive-text_secondary)",
        "text_tertiary": "var(--positive-text_tertiary)",
        "bg_base": "var(--positive-bg_base)",
        "bg_base_hover": "var(--positive-bg_base_hover)",
        "bg_base_active": "var(--positive-bg_base_active)",
        "bg_accent": "var(--positive-bg_accent)",
        "bg_accent_hover": "var(--positive-bg_accent_hover)",
        "bg_accent_active": "var(--positive-bg_accent_active)",
        "bg_tint": "var(--positive-bg_tint)",
        "icon_default": "var(--positive-icon_default)",
        "icon_inverted": "var(--positive-icon_inverted)",
        "border_base": "var(--positive-border_base)",
        "border_base_hover": "var(--positive-border_base_hover)",
        "border_accent": "var(--positive-border_accent)",
        "border_accent_inverted": "var(--positive-border_accent_inverted)"
    },
    "caution": {
        "text_primary": "var(--caution-text_primary)",
        "text_secondary": "var(--caution-text_secondary)",
        "text_tertiary": "var(--caution-text_tertiary)",
        "bg_base": "var(--caution-bg_base)",
        "bg_base_hover": "var(--caution-bg_base_hover)",
        "bg_base_active": "var(--caution-bg_base_active)",
        "bg_accent": "var(--caution-bg_accent)",
        "bg_accent_hover": "var(--caution-bg_accent_hover)",
        "bg_accent_active": "var(--caution-bg_accent_active)",
        "bg_tint": "var(--caution-bg_tint)",
        "icon_default": "var(--caution-icon_default)",
        "icon_inverted": "var(--caution-icon_inverted)",
        "border_base": "var(--caution-border_base)",
        "border_base_hover": "var(--caution-border_base_hover)",
        "border_accent": "var(--caution-border_accent)",
        "border_accent_inverted": "var(--caution-border_accent_inverted)"
    },
    "critical": {
        "text_primary": "var(--critical-text_primary)",
        "text_secondary": "var(--critical-text_secondary)",
        "text_tertiary": "var(--critical-text_tertiary)",
        "bg_base": "var(--critical-bg_base)",
        "bg_base_hover": "var(--critical-bg_base_hover)",
        "bg_base_active": "var(--critical-bg_base_active)",
        "bg_accent": "var(--critical-bg_accent)",
        "bg_accent_hover": "var(--critical-bg_accent_hover)",
        "bg_accent_active": "var(--critical-bg_accent_active)",
        "bg_tint": "var(--critical-bg_tint)",
        "icon_default": "var(--critical-icon_default)",
        "icon_inverted": "var(--critical-icon_inverted)",
        "border_base": "var(--critical-border_base)",
        "border_base_hover": "var(--critical-border_base_hover)",
        "border_accent": "var(--critical-border_accent)",
        "border_accent_inverted": "var(--critical-border_accent_inverted)"
    }
}
```

src/theme/lib/theme/color/defaults.ts
```diff
@@ -249,64 +259,6 @@ export const defaultOpts: ThemeColorBuilderOpts = {
     }
   },
 
-  button: ({base, mode, muted, solid}) => {
```
pedrobonamin, line 252: Aiming for simplicity

src/primitives/card/card.tsx
```diff
@@ -65,24 +67,26 @@ export const Card = forwardRef(function Card(
...
   const as = isValidElementType(asProp) ? asProp : 'div'
-  const rootTheme = useRootTheme()
-  const tone = toneProp === 'inherit' ? rootTheme.tone : toneProp
+  const toneContext = useToneContext()
+  const tone = toneProp === 'inherit' ? toneContext.tone : toneProp
+  const scheme = schemeProp ?? toneContext.scheme
 
   return (
-    <ThemeColorProvider scheme={scheme} tone={tone}>
+    <ToneProvider scheme={scheme} tone={tone}>
```
pedrobonamin, line 83: <ThemeColorProvider> is not needed anymore, we need this toneProvider as some inner elements need to know which tone and scheme the card has.
Also, when nesting cards we need to preserve the nearest card tone and scheme

src/primitives/textInput/textInput.tsx
```diff
@@ -339,22 +341,20 @@ export const TextInput = forwardRef(function TextInput(
   )
 
   return (
-    <Root data-ui="TextInput" tone={rootTheme.tone}>
+    <Root data-ui="TextInput" tone={tone}>
```
pedrobonamin, line 344: This is returning a new `<Card /> `component, do we need this? 
I assume it's one of the most used components, and it's creating a whole new context of css vars and of toneContext without needing it?

How can we test if it's really needed? Is there a use case for this? 
I assume we could return just a div element with the necessary styles. 
cc: @mariuslundgard @robinpyon

src/styles/colorVars/colorVarsStyle.ts
```diff
@@ -1,43 +1,144 @@
-import {ThemeColorBase, ThemeColorGenericState} from '../../theme'
+import {cssVars} from '../../theme'
```
pedrobonamin, line 1: Not sure about this, it's worth discussing it.
This is used by the `<Tree />` , `<Selectable />` and ` <Card /> ` it adds variables to the context of each of this elements. 
It falls outside of the logic of how variables are handled with the new implementation, worth discussing the correct location of this file.

It's basically a rewriting of variables to keep consistency with the previous implementaiton.
Also, some of this variables `mutableCardVariables` are updated by other components outisde of the card.

src/theme/studioTheme/tints.ts
```diff
@@ -1,67 +1,171 @@
 /**
```
pedrobonamin, line 1: Should this file belong here? Or should this be moved alongside the variables files?

src/primitives/avatar/avatarCounter.tsx
```diff
@@ -32,11 +33,11 @@ function _avatarCounterBaseStyle(props: ThemeProps) {
     box-sizing: border-box;
     user-select: none;
     color: inherit;
-    color: var(--card-fg-color);
-    background: var(--card-bg-color);
+    color: ${cssVars.mutable['fg-color']};
```
pedrobonamin, line 36: This component is a good example of the usage of mutable and non mutable variables.
In this case, the mutable variables are possibly updated by parents higher in the tree, like the color of the avatar counter. And some other variables are only updated by the card or will be received from the root.

src/primitives/textInput/textInput.tsx
```diff
@@ -163,12 +165,14 @@ export const TextInput = forwardRef(function TextInput(
     customValidity,
     type = 'text',
     weight,
+    tone: toneProp = 'default',
```
pedrobonamin, line 168: New possibility in inputs, to set the tone directly to them, avoiding the need to wrap it with a card

src/theme/lib/color-fns/types.ts
```diff
@@ -5,6 +5,17 @@ export interface RGB {
   r: number
   g: number
   b: number
+  a?: undefined
+}
+
+/**
+ * @internal
+ */
+export interface RGBA {
```
pedrobonamin, line 14: This is introduced because colors can have opacity, and when mixing them we need to remove the opacity of the hex, do the mix and then return the opacity added after the mix.

src/theme/lib/theme/color/cssVariables/createCssVars.ts
```diff
@@ -0,0 +1,55 @@
...
+import {ThemeColorName, ThemeColorSchemeKey} from '../types'
+import {createSpotVars, spotCssVariables} from './spot'
+import {createSyntaxVariables, syntaxCssVariables} from './syntax'
+import {createTonesVariables, tonesCssVariables} from './tones'
+
+/**
+ *
+ * @beta
+ */
+export const createCssVars = (
```
pedrobonamin, line 13: This creates all the css variables that are going to be available in the root and then recreated again in the cards if needed. 
It will do a mixing if it receives a `defaultTone` different to `default`.

src/theme/lib/theme/color/types.ts
```diff
@@ -35,38 +27,10 @@ export type ThemeColorToneKey =
...
-  critical: ThemeColor
-}
-
 /**
  * @public
  */
 export interface ThemeColorSchemes {
-  dark: ThemeColorScheme
-  light: ThemeColorScheme
+  // Could we update this to only return Record<ThemeColorName ,  keyof typeof hues> ? Then the function that assigns the color would search the ColorTint
```
pedrobonamin, line 34: Open question to discuss:
Do we want to allow users to import any color? Or only sanity/color colors?
If we wnat to restrict to sanity colors, we can update this to receive only the name of the colors, and then handle the rest internally.

src/theme/themeProvider.tsx
```diff
@@ -55,7 +67,11 @@ export function ThemeProvider(props: ThemeProviderProps): React.ReactElement {
 
   return (
     <ThemeContext.Provider value={value}>
-      <StyledThemeProvider theme={theme}>{children}</StyledThemeProvider>
+      <StyledThemeProvider theme={theme}>
```
pedrobonamin, line 70: This could be removed if we deprecate the possibility to change through the theme config  the rest of options (sizes, radius, shadows, etc)

src/theme/lib/theme/color/types.ts
robinpyon: Are you able to contextualise this a bit more? Are you talking about users defining custom colours in a future theme api, or maintainers updating default studio tints / colour values?

src/theme/studioTheme/theme.ts
```diff
@@ -21,7 +21,13 @@ export const studioTheme: RootTheme = {
   card: {
     focusRing: {offset: -1, width: 1},
   },
-  color,
+  color: {
+    tones: tones,
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore This is necessary until workshop is updated.
+    base: {bg: '#fffff'},
```
robinpyon, line 29: **question(non-blocking):** can you elaborate as to why workshop needs this to be set? Is it still necessary given that we're moving to storybook?

src/theme/studioTheme/tints.ts
robinpyon: `colorKeys` is agnostic to the studio's default them and used elsewhere, so I think it makes sense to move it

robinpyon (approved):
Thanks @pedrobonamin, this is a huge effort. Leaning on CSS variables more is definitely a step in the right direction.

We should look to formulate a more rigid approach to how we name these and think how we can better surface the concept of "mutable" variables, but this is a fantastic starting point for us. Great work!

src/theme/lib/theme/color/types.ts
pedrobonamin: Right now our tones object looks like this:
```
const tones = {
  default: {
    "50": { title: "Gray 50", hex: "#f9f9fa" },
    // tints in between
    "950": { title: "Gray 950", hex: "#17191c" },
  },
  transparent: {
    "50": { title: "Gray 50", hex: "#f9f9fa" },
    // tints in between
    "950": { title: "Gray 950", hex: "#17191c" },
  },
  primary: {
    //  gray tints from 50 to 950 from sanity ui}
  },
  positive: {
    // cyan tints from 50 to 950 from sanity ui}
  },
  caution: {
    // yellow tints from 50 to 950 from sanity ui}
  },
  critical: {
    // red tints from 50 to 950 from sanity ui}
  },
};

```

If we only want to allow users to use colors from `@sanity/ui` we could reduce tones to just be:
```
export const tones: Record<ThemeColorName, ColorHueKey> = {
  default: 'gray',
  transparent: 'gray',
  primary: 'gray',
  positive: 'cyan',
  caution: 'yellow',
  critical: 'red',
} 
```
This change will make it simpler to update the colors from a theming perspective, in which the user only selects the 6 necessary colors.

src/theme/studioTheme/theme.ts
pedrobonamin: The workshop imported in studio ui uses the theme from sanity/ui, and it  uses the base bg color in `<GlobalStyle />` https://github.com/sanity-io/ui-workshop/blob/main/src/core/GlobalStyle.ts.
If the theme that we are exporting here doesn't have this object, when running workshop it throws the following error:
Screenshot 2023-10-17 at 17 37 41

src/theme/lib/theme/color/types.ts
robinpyon: Thanks for clarifying! Suggest that we leave as is for now until we land the new theming API surface

src/primitives/textInput/textInput.tsx
pedrobonamin: I think it's not necessary anymore, anyways would prefer to handle it in a different ticket, as the input has some other uses of card in `Prefix` `Suffix` and `RightCard` all creating new cards.
Would be great to revisit all of those uses and decide if we can remove them (probably we can) but not sure what are all the uses cases involved into this.

What do you think?

robinpyon: Agreed!

src/theme/lib/theme/color/types.ts
pedrobonamin: Cool, makes sense, just wanted to highlight that we have a possible change to simplify things 😄

PR #1155, opened 2023-10-04
feat(dialog): update styles and add corresponding stories

ninaandal:
Adds a storybook story for the `dialog` component. 
Also edits the borders for the content in the dialog component from always showing, to show conditionally on scroll


pedrobonamin (approved):

robinpyon (approved):
Thanks @ninaandal! Just added a sneaky commit to remove unnecessary setState calls – looks great!

PR #1156, opened 2023-10-04
refactor(storybook): clean up dialog stories

robinpyon:
(Comments and feedback have been left in code comments!)

### DIFF
--- a/stories/components/Dialog.stories.tsx
+++ b/stories/components/Dialog.stories.tsx
@@ -1,9 +1,10 @@
-import {ArrowDownIcon, ArrowUpIcon, CloseIcon, TrashIcon} from '@sanity/icons'
-import type {Meta, StoryObj} from '@storybook/react'
+/* eslint-disable react-hooks/rules-of-hooks */
+import {ArrowDownIcon, ArrowUpIcon} from '@sanity/icons'
+import type {Meta, StoryFn, StoryObj} from '@storybook/react'
 import {useCallback, useState} from 'react'
 import {Dialog} from '../../src/components'
 import {Box, Button, Card, Inline, Stack, Text} from '../../src/primitives'
-import {LayerProvider} from '../../src/utils'
+import {BoundaryElementProvider, PortalProvider} from '../../src/utils'
 import {
   getContainerWidthControls,
   getPositionControls,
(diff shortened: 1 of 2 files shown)


stories/components/Dialog.stories.tsx
```diff
@@ -13,22 +14,72 @@ import {
...
+    ),
+    footer: (
+      <Card padding={4}>
+        <Inline>
+          <Text weight="medium">Dialog footer</Text>
+        </Inline>
+      </Card>
+    ),
+    header: 'Dialog header',
+  },
```
robinpyon, line 43: **suggestion:** define args that can serve as sensible defaults across all stories (individual stories can always null these out as necessary)

stories/components/Dialog.stories.tsx
```diff
@@ -13,22 +14,72 @@ import {
...
     paddingLeft: getSpaceControls(),
     paddingRight: getSpaceControls(),
     paddingTop: getSpaceControls(),
     paddingX: getSpaceControls(),
     paddingY: getSpaceControls(),
-    width: getContainerWidthControls(),
-    cardRadius: getRadiusControls(),
     position: getPositionControls(),
-    cardShadow: getShadowControls(),
-    footer: {control: {disable: true}},
```
robinpyon, line 29: **suggestion:** it may be worth showing `footer` controls – even if it receives a `ReactNode`, surfacing this in docs may be valuable

stories/components/Dialog.stories.tsx
```diff
@@ -13,22 +14,72 @@ import {
...
+    },
+  ],
+  parameters: {
+    docs: {
+      story: {
+        inline: false,
+        iframeHeight: 400,
+      },
+    },
+  },
```
robinpyon, line 82: **suggestion:** Storybook specific config to limit stories on the doc page to a specific height (which is handy when we consider the _Positioning_ story)

stories/components/Dialog.stories.tsx
robinpyon: Forgot to mention, `__unstable_autoFocus: true` was added to ensure that we prevent autofocus doesn't kick in on the docs page (which would otherwise focus the last dialog on this page)

ninaandal (approved):
This is great, thank you so much for explanations and input 🤌

PR #1157, opened 2023-10-04
feat: update KBD primitive style, minor storybook cleanup

robinpyon:
### Description

This PR updates the `<KBD>` primitive, rendering its background with `--card-bg2-color`. 

This is a new exported color which is based off `color.bg2`. I believe this is the idiomatic approach that other primitives follow (using css vars in their corresponding `<Root>` component) – does this track?

**Storybook additions:**
- Removed `radiusBuilder` in-lieu of a new `rowBuilder`, which is similar in signature to the current matrix builder. 
- Refactored a few stories to the use the above


pedrobonamin (approved):
Thanks for this! 
Will it be worth having this as a Studio component? I think of having a wrapper that receives an array of `keys` and it returns one kbd component for each key. With the proper spacing applied.

robinpyon:
Thanks @pedrobonamin! 

> I think of having a wrapper that receives an array of keys and it returns one kbd component for each key. With the proper spacing applied.

That exists in `<Hotkeys>`!

PR #1161, opened 2023-10-09
fix: update dialog header appearance, update on element resize + add story

robinpyon:
(Comments and feedback have been left in code comments)

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -14,6 +14,7 @@ import {responsivePaddingStyle, ResponsivePaddingStyleProps} from '../../styles/
 import {ThemeColorSchemeKey, useTheme} from '../../theme'
 import {DialogPosition, Radius} from '../../types'
 import {Layer, LayerProps, Portal, useBoundaryElement, useLayer, usePortal} from '../../utils'
+import {BORDER_OFFSET_X} from './constants'
 import {
   dialogStyle,
   responsiveDialogPositionStyle,
@@ -72,12 +73,12 @@ interface DialogCardProps extends ResponsiveWidthProps {
 }
 
 interface DialogHeaderProps extends CardProps {
-  hasScrolledFromTop: boolean
-  isContentScrollable: boolean
+  $hasScrolledFromTop: boolean
+  $isContentScrollable: boolean
 }
 interface DialogFooterProps extends CardProps {
-  hasScrolledToBottom: boolean
-  isContentScrollable: boolean
+  $hasScrolledToBottom: boolean
+  $isContentScrollable: boolean
 }
 
 function isTargetWithinScope(
@@ -135,15 +136,13 @@ const DialogHeader = styled(Card)<DialogHeaderProps>`
     content: '';
     display: block;
     position: absolute;
-    left: 0;
-    right: 0;
+    left: ${BORDER_OFFSET_X}px;
+    right: ${BORDER_OFFSET_X}px;
     bottom: -1px;
-    border-bottom: ${(props) =>
-      props.isContentScrollable &&
-      props.hasScrolledFromTop &&
-      '1px solid var(--card-border-color)'};
-    width: calc(100% - 24px);
-    margin-left: 12px;
+    border-bottom: 1px solid var(--card-border-color);
+    width: auto;
+    opacity: ${(props) => (props.$isContentScrollable && props.$hasScrolledFromTop ? 1 : 0)};
+    transition: opacity 200ms ease-in;
   }
 `
 
@@ -157,12 +156,19 @@ const DialogContent = styled(Box)`
 const DialogFooter = styled(Box)<DialogFooterProps>`
   position: relative;
   z-index: 3;
-  border-top: ${(props) =>
-    !props.hasScrolledToBottom &&
-    props.isContentScrollable &&
-    '1px solid var(--card-border-color)'};
-  width: calc(100% - 24px);
-  margin-left: 12px;
+
+  &:before {
+    content: '';
+    display: block;
+    position: absolute;
+    left: ${BORDER_OFFSET_X}px;
+    right: ${BORDER_OFFSET_X}px;
+    top: -1px;
+    border-top: 1px solid var(--card-border-color);
+    width: auto;
+    opacity: ${(props) => (props.$isContentScrollable && !props.$hasScrolledToBottom ? 1 : 0)};
+    transition: opacity 200ms ease-in;
+  }
 `
 
 const DialogCard = forwardRef(function DialogCard(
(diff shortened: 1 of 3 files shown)


src/components/dialog/dialog.tsx
```diff
@@ -135,15 +136,13 @@ const DialogHeader = styled(Card)<DialogHeaderProps>`
...
-    border-bottom: ${(props) =>
-      props.isContentScrollable &&
-      props.hasScrolledFromTop &&
-      '1px solid var(--card-border-color)'};
-    width: calc(100% - 24px);
-    margin-left: 12px;
+    border-bottom: 1px solid var(--card-border-color);
+    width: auto;
+    opacity: ${(props) => (props.isContentScrollable && props.hasScrolledFromTop ? 1 : 0)};
+    transition: opacity 200ms ease-in;
```
robinpyon, line 145: **suggestions:**

- remove magic numbers and extract into a separate const, following the pattern used by other primitives / components (in case we want to change this later)
- always render the border but instead we vary on opacity to [prevent repaints ](https://web.dev/simplify-paint-complexity-and-reduce-paint-areas)([also relevant](https://web.dev/animations-overview/#pipeline))
- added basic transition on `opacity` – see above for why we do this on opacity instead of border
- since we're positioning absolutely, it's probably a little more readable to set offsets on `left` and `right` values here (rather than setting them to 0, then adding a margin + calculating `width`)

src/components/dialog/dialog.tsx
```diff
@@ -157,12 +156,19 @@ const DialogContent = styled(Box)`
...
+    display: block;
+    position: absolute;
+    left: ${BORDER_OFFSET_X}px;
+    right: ${BORDER_OFFSET_X}px;
+    top: -1px;
+    border-top: 1px solid var(--card-border-color);
+    width: auto;
+    opacity: ${(props) => (props.isContentScrollable && !props.hasScrolledToBottom ? 1 : 0)};
+    transition: opacity 200ms ease-in;
+  }
```
robinpyon, line 171: **suggestion:** use `:before` pseudo element (similar to `<DialogHeader>`)

src/components/dialog/dialog.tsx
```diff
@@ -289,8 +295,9 @@ const DialogCard = forwardRef(function DialogCard(
     }
   }, [])
 
+  // Update scroll header + footer border state on scroll and content mutation
   useEffect(() => {
-    const handleResize = () => {
+    const handleUpdate = () => {
```
robinpyon, line 300: **suggestion:** rename to `handleUpdate` since this is triggered by mutation updates too

src/components/dialog/dialog.tsx
```diff
@@ -135,15 +136,13 @@ const DialogHeader = styled(Card)<DialogHeaderProps>`
...
-    border-bottom: ${(props) =>
-      props.isContentScrollable &&
-      props.hasScrolledFromTop &&
-      '1px solid var(--card-border-color)'};
-    width: calc(100% - 24px);
-    margin-left: 12px;
+    border-bottom: 1px solid var(--card-border-color);
+    width: auto;
+    opacity: ${(props) => (props.$isContentScrollable && props.$hasScrolledFromTop ? 1 : 0)};
+    transition: opacity 200ms ease-in;
```
robinpyon, line 145: **suggestions:**

- remove magic numbers and extract into a separate const, following the pattern used by other primitives / components (in case we want to change this later)
- always render the border but instead we vary on opacity to [prevent repaints ](https://web.dev/simplify-paint-complexity-and-reduce-paint-areas)([also relevant](https://web.dev/animations-overview/#pipeline))
- added basic transition on `opacity` – see above for why we do this on opacity instead of border
- since we're positioning absolutely, it's probably a little more readable to set offsets on `left` and `right` values here (rather than setting them to 0, then adding a margin + calculating `width`)

src/components/dialog/dialog.tsx
```diff
@@ -72,12 +73,12 @@ interface DialogCardProps extends ResponsiveWidthProps {
 }
 
 interface DialogHeaderProps extends CardProps {
-  hasScrolledFromTop: boolean
-  isContentScrollable: boolean
+  $hasScrolledFromTop: boolean
+  $isContentScrollable: boolean
```
robinpyon, line 77: **suggestion:** prefix styled-component props with `$` – this not only helps remove React warnings in the console (`React does not recognize the `X` prop on a DOM element.`) but also helps signify these are props on a styled component, which is a convention Sanity UI uses quite extensively.

src/components/dialog/dialog.tsx
```diff
@@ -157,12 +156,19 @@ const DialogContent = styled(Box)`
...
+    display: block;
+    position: absolute;
+    left: ${BORDER_OFFSET_X}px;
+    right: ${BORDER_OFFSET_X}px;
+    top: -1px;
+    border-top: 1px solid var(--card-border-color);
+    width: auto;
+    opacity: ${(props) => (props.$isContentScrollable && !props.$hasScrolledToBottom ? 1 : 0)};
+    transition: opacity 200ms ease-in;
+  }
```
robinpyon, line 171: **suggestion:** wrap this in a `:before` pseudo element (similar to `<DialogHeader>`). This is also required to allow us to change opacity on this pseudo element only (rather than the entire header / footer)

src/components/dialog/dialog.tsx
```diff
@@ -290,39 +296,30 @@ const DialogCard = forwardRef(function DialogCard(
...
-        setHasScrolledFromTop(localContentRef.current.scrollTop > 0)
-        setHasScrolledToBottom(
-          localContentRef.current.scrollTop >=
-            localContentRef.current.scrollHeight - localContentRef.current.clientHeight,
-        )
-      }
-    }
+    const resizeObserver = new ResizeObserver(handleContentUpdate)
 
-    window.addEventListener('resize', handleResize)
```
robinpyon, line 306: **suggestion:** rather than listening to resize events on the entire _window_, we should instead only listen to resize events on the target element with a [ResizeObserver](https://web.dev/resize-observer/). 

Doing so means that we can also handle cases where content is added dynamically – for example, if dialog content doesn't originally overflow, but content is added which then adds a scrollbar. Listening to window events won't handle this scenario, and a separate story has been created to demonstrate this.

src/components/dialog/dialog.tsx
```diff
@@ -290,39 +296,30 @@ const DialogCard = forwardRef(function DialogCard(
...
-              isContentScrollable={isContentScrollable}
-              hasScrolledFromTop={hasScrolledFromTop}
+              $isContentScrollable={isContentScrollable}
+              $hasScrolledFromTop={hasScrolledFromTop}
             >
               <Flex>
                 <Box flex={1} padding={4}>
                   {header && (
-                    <Text id={labelId} weight="semibold">
+                    <Text id={labelId} size={1} weight="medium">
```
robinpyon, line 322: **suggestion:** change default dialog header styles to match new layout

PR #1164, opened 2023-10-16
fix(theme): set root CSS variables directly in the document stylesheet

robinpyon:
### Description

This PR declares css variables directly in the document head (via styled-components' `createGlobalStyle`)

**Storybook changes:**
- Added a few storybook stories highlighting css tone vars and spot colours
- Minor refactoring: now pass `scheme` where possible and work around the assumption that there may only been two scheme types (there could be more in future)
- Minor change to the main storybook decorator to better accommodate background rendering within the docs page


pedrobonamin (approved):
Thanks for this changes!

PR #1165, opened 2023-10-16
chore(css-vars): reduce the number of css variables available in tints.

pedrobonamin:
Divides the css variables according to tones related and default ones.
Some variables, are only accessed by the default tone, so there is no need to create them into each of the tones.
This change, reduces the variables from 200 to 126.


pedrobonamin:
Closing as the changes for css variables will be reverted

PR #1166, opened 2023-10-17
feat(menu): update menu and menu items styles

pedrobonamin:
Updates menu and menu item styles to match facelift. 
Refactors menu item to support new props:
* `subText`
* `badgeText`

https://sanity-ui-storybook-git-edx-484-menu.sanity.build/?path=/story/components-menu--menu-items-variants&globals=theme:light

## Examples: 

* Left - Menu items enabled.
* Right - Menu items disabled.
### Light mode
Screenshot 2023-10-17 at 13 16 10

### Dark mode
Screenshot 2023-10-17 at 13 16 16


src/primitives/badge/badge.tsx
```diff
@@ -51,7 +51,9 @@ export const Badge = forwardRef(function Badge(
       padding={useArrayProp(padding)}
       ref={ref}
     >
-      <Label size={fontSize}>{children}</Label>
+      <Text size={fontSize} weight="medium">
```
pedrobonamin, line 54: Replaced for text, as badge supports lowercase according to designs.

pedrobonamin: Thinking if we should we aim to deprecate `<Label />` as it seems to be the same as `<Text />` but forcing an uppercase. 
cc: @robinpyon

robinpyon: I think it makes sense to, given the old `<Label>` no longer appears to be present. Can you confirm @mikolajdobrucki?

src/components/menu/menuItem.tsx
```diff
@@ -31,6 +31,8 @@ export interface MenuItemProps extends ResponsivePaddingProps, ResponsiveRadiusP
   space?: number | number[]
   text?: React.ReactNode
   tone?: ThemeColorToneKey
+  subText?: React.ReactNode
+  badgeText?: React.ReactNode
```
robinpyon, line 35: **suggestion:** I would consider removing these – however, I believe exposing these props would make more sense in _Studio_ UI.

I'd think otherwise if `text` only accepted a `string`, but it currently accepts a `ReactNode` and our current docs has examples which demonstrates how you can create a similar layout ([example](https://www.sanity.io/ui/docs/component/Menu#example-with-custom-menu-items))

However in Studio UI we can be way more restrictive – and even take it a step further, only allowing basic `string` values . WDYT?

src/primitives/_selectable/style.ts
robinpyon, line 1: **question:** what's the rationale for applying hover styles to both `active` and `selected` states here? These changes mean that it's not possible to disambiguate between the 'selected' menu item and the one that's being currently hovered.

To better illustrate, try visit [this example](https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJyzccqvUChITEnJzEu3rTapteNSULBxTixKUShKTMksLbatNq5VKM5ITMkvt602AksDFfim5pVCmFCOZ0lqLpSvoJCWn1cSnFmValsdbaSjAETGOgomsbVw%2BZLUihJbJbfMouIShfyCksz8PCWonD6lhganJufnpVDb1JCMzCIchtrow8LCRh8UbHZcNvrAILXjRAYA54VbAw%3D%3D&title=Menu+example&description=A+basic+example+of+using+the+Menu+component+in+Sanity+UI.) and click any menu item and then press the up or down arrows – you'll be able to see the item that is 'selected' whilst having a hover state on the previously clicked button

src/components/menu/menuItem.tsx
pedrobonamin: Discussed by slack, agreed in removing this and moving into Studio UI, thanks @robinpyon .

src/primitives/badge/badge.tsx
pedrobonamin: Created a ticket to track this.

src/primitives/_selectable/style.ts
pedrobonamin: Hey @robinpyon  In the figma file the selected and hover states are the same that's the reason they are repeated.
With the redesign, the tint used for the selected state background is `50`  so we don't have a way to set a lighter color in hover like in the example.

robinpyon (approved):

PR #1168, opened 2023-10-18
feat(hotkeys): platform aware keys (alt => option, option => alt)

rexxars:
When rendering hotkeys such as `Ctrl + Alt + I`, it would be ideal if we showed the platform-specific hotkeys, eg `Option` instead of `Alt` on Windows/Linux. Similarly, when specifying `Ctrl + Option + I`, it should adjust according to platform.

I wasn't quite sure if we should also add "Opt" as a case, but decided against it for now - open to input.

Theoretically a user can of course have a Mac keyboard connected to a Windows/Linux computer and vice versa, but my gut feeling that it is an edge case not worth considering given we can't actually detect the situation.

### DIFF
--- a/src/components/hotkeys/__workshop__/plain.tsx
+++ b/src/components/hotkeys/__workshop__/plain.tsx
@@ -1,9 +1,21 @@
-import {Flex, Hotkeys} from '@sanity/ui'
+import {Flex, Hotkeys, Label, Stack} from '@sanity/ui'
 
 export default function PlainStory() {
   return (
     <Flex align="center" height="fill" justify="center">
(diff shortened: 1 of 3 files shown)


rexxars:
Starting to doubt myself on whether or not this is a good idea for `@sanity/ui` - maybe this belongs as a `platformAwareKeys('Ctrl+Alt+I')` function in the `sanity` package instead 🤔

rexxars:
After talking this through, I think this is the wrong place for this logic. The UI framework should present thing without too much magic - you may want to use it to display both the windows and mac keyboard shortcuts, for instance, which this will prevent you from doing.

PR #1169, opened 2023-10-19
fix(dialog): update dialog design

pedrobonamin:
- Updates dialog default padding values in body and footer.
- Updates header padding and structure. 
- Added story with example for the delete document.
Screenshot 2023-10-19 at 17 34 01

### DIFF
--- a/src/theme/studioTheme/theme.ts
+++ b/src/theme/studioTheme/theme.ts
@@ -28,7 +28,7 @@ export const studioTheme: RootTheme = {
     // @ts-ignore This is necessary until workshop is updated.
     base: {bg: '#fffff'},
   },
-  container: [360, 600, 960, 1280, 1600, 1920],
+  container: [320, 640, 960, 1280, 1600, 1920],
   focusRing: {
     offset: 1,
     width: 2,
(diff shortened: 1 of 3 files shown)


src/theme/studioTheme/theme.ts
```diff
@@ -28,7 +28,7 @@ export const studioTheme: RootTheme = {
     // @ts-ignore This is necessary until workshop is updated.
     base: {bg: '#fffff'},
   },
-  container: [320, 640, 960, 1280, 1600, 1920],
+  container: [360, 600, 960, 1280, 1600, 1920],
```
robinpyon, line 31: **question:** wondering if this change is necessary? is it required to fix a behaviour of current dialog components?

pedrobonamin: The dialogs have different widths in the designs that don't match with this values. 
Asked in figma to @mikolajdobrucki to confirm the change.

robinpyon: Thanks! In this instance, I'd suggest we err on the side of caution and keep `container` values as is, as they're used in multiple places (popovers, containers themselves, etc)

robinpyon (approved):

PR #1171, opened 2023-10-25
fix(tooltip): update tooltip position middleware

pedrobonamin:
Updates how the tooltip position middleware works.

- Updates the Portal to be created into `body` when using `portal=true` in tooltips. Allow for tooltip to show on top of panes
- Removes the size function, as it was adding a maxHeight and width to the tooltip element, making it behave in unexpected ways when the available space was smaller than the needed space.
- Removes the `mainAxis=false` from the flip, as it produces tooltips to not flip in the main axis, so if you have a tooltip with `position="top"` and it doesn't fit in the top position, it won't flip as it's disabled by this flag. https://floating-ui.com/docs/flip#combining-with-shift
- Removes the contextBoundary when flipping the tooltip, as it made the library understand that it has more space to show the elements than the actual available space, the contextBoundary receives the entire pain.

### Before
Screenshot 2023-10-25 at 14 16 55

Screenshot 2023-10-25 at 14 16 55

### After
Screenshot 2023-10-25 at 14 17 17

Screenshot 2023-10-25 at 14 21 41

### Some other cases:

https://github.com/sanity-io/ui/assets/46196328/9b18a69d-f772-4640-b14c-c67e779a66d5

In comments branch the fallbackPlacement prop needs to be removed. Will add file reference and notify the team about it.

### DIFF
--- a/src/primitives/tooltip/__workshop__/index.ts
+++ b/src/primitives/tooltip/__workshop__/index.ts
@@ -15,5 +15,10 @@ export default defineScope({
       title: 'Resizable Boundary',
       component: lazy(() => import('./resizableBoundary')),
     },
+    {
+      name: 'overflowBoundary',
+      title: 'Overflowing Boundary',
+      component: lazy(() => import('./overflowingBoundary')),
+    },
   ],
 })
--- a/src/primitives/tooltip/__workshop__/index.ts
+++ b/src/primitives/tooltip/__workshop__/index.ts
@@ -10,5 +10,10 @@ export default defineScope({
       title: 'Props',
       component: lazy(() => import('./props')),
     },
+    {
+      name: 'resizableBoundary',
+      title: 'Resizable Boundary',
+      component: lazy(() => import('./resizableBoundary')),
+    },
   ],
 })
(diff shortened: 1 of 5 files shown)


robinpyon:
Thanks @pedrobonamin – I've noticed this causes a regression with tooltips causing unwanted overflow (which I believe the original commit sought to resolve). I've added a simple workshop story which demonstrates this (`tooltip/resizableBoundary`).

pedrobonamin:
> Thanks @pedrobonamin – I've noticed this causes a regression with tooltips causing unwanted overflow (which I believe the original commit sought to resolve). I've added a simple workshop story which demonstrates this (`tooltip/resizableBoundary`).

Thanks @robinpyon I think in that case we should only keep the maxWidth, as the maxHeight is the one causing the issue of the arrow being incorrectly positioned. 
Tried in the story and it looks correct, also tried in the studio in the cases that were reported and they all look ok.

ninaandal (approved):
Great addition with the story @robinpyon, makes it clear why we added the `maxWidth` in the first place! 
But this looks great now @pedrobonamin 🎉

robinpyon (approved):
Thanks @pedrobonamin!

I made a few minor changes to bring it a bit more inline with the existing popover component, added a failing test case (when constrained within a boundary element with loads of text) and removed the `size` middleware from `floating-ui`, which I believe isn't needed when dealing with tooltips.

PR #1172, opened 2023-10-26
chore(css-variables): revert changes for css variables

pedrobonamin:
As discussed with the team, this PR reverts the implementation of css variables.
It restores the previous color theming through styled components, aiming for backwards compatibility.

It doesn't remove some stories that will be useful to validate future color changes as `tones`.
It doesn't revert the changes to tints, it just moves it to the studioTheme folder given that possibly it will be used for the color updates. 

What to review:
Only changes to the color theming should be reverted.


src/primitives/card/card.tsx
```diff
@@ -55,29 +47,6 @@ const Root = styled(Box)<
     ResponsiveShadowStyleProps
 >(responsiveBorderStyle, responsiveRadiusStyle, responsiveShadowStyle, cardStyle)
 
-const Wrapper = ({
```
pedrobonamin, line 58: I think bringing back later something like this will be useful to avoid having so many theme provider context in the studio. 
Just adding this to think about it

robinpyon (approved):
Thanks @pedrobonamin!

Minor one – given that this is a bit of a reset, should we consider reverting `tones` in `studioTheme/color` too? (e.g. primary to blue – positive to green)

PR #1175, opened 2023-10-31
feat(buttons): add support for width=fill prop

pedrobonamin:
Adds new prop `width?: "fill"` to buttons. Supporting full width buttons

### DIFF
--- a/src/primitives/button/button.tsx
+++ b/src/primitives/button/button.tsx
@@ -5,7 +5,7 @@ import {useArrayProp} from '../../hooks'
 import {ThemeProps} from '../../styles'
 import {responsiveRadiusStyle, ResponsiveRadiusStyleProps} from '../../styles/internal'
 import {useTheme} from '../../theme'
-import {ButtonMode, ButtonTextAlign, ButtonTone, FlexJustify} from '../../types'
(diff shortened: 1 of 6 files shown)


src/primitives/button/button.tsx
```diff
@@ -34,10 +34,11 @@ export interface ButtonProps extends ResponsivePaddingProps, ResponsiveRadiusPro
   text?: React.ReactNode
   tone?: ButtonTone
   type?: 'button' | 'reset' | 'submit'
+  width?: 'fill'
 }
 
 const Root = styled.button<
-  {$mode: ButtonMode; $tone: ButtonTone} & ResponsiveRadiusStyleProps & ThemeProps
+  {$mode: ButtonMode; $tone: ButtonTone; $width?: 'fill'} & ResponsiveRadiusStyleProps & ThemeProps
```
robinpyon, line 41: **question:** what do you think about creating a separate `ButtonWidth` type? Whilst it only supports one value right now, it means we can avoid using a hard-coded value in [button/styles.ts](https://github.com/sanity-io/ui/pull/1175/files#diff-5bb0a23cf7090fc7b683864f58573e31f1f1c0667bff98d66af509b3c0d92eb4)

robinpyon (approved):
Thanks @pedrobonamin!

PR #1181, opened 2023-11-07
refactor: theme builder components

mariuslundgard:


src/components/toast/toast.tsx
```diff
@@ -56,14 +57,18 @@ export function Toast(
...
     >
       <Flex align="flex-start">
         <TextBox flex={1} padding={3}>
           <Stack space={3}>
-            {title && <Text weight="semibold">{title}</Text>}
+            {title && (
+              <Text size={1} weight="medium">
+                {title}
+              </Text>
+            )}
```
robinpyon, line 71: It's incredible how big a difference this makes – these new toasts look great.

src/primitives/badge/styles.ts
robinpyon, line 1: Safe to say that `mode="outline"` will be deprecated and we'll add a new `ghost` mode? (i.e. badges won't have an outlined style?)

src/components/dialog/dialog.tsx
```diff
@@ -311,7 +311,7 @@ const DialogCard = forwardRef(function DialogCard(
               $hasScrolledFromTop={hasScrolledFromTop}
             >
               <Flex align="center">
-                <Box flex={1} margin={1} padding={4}>
+                <Box flex={1} padding={4}>
```
robinpyon, line 314: This is an interesting one – AFAIK, the spacing of the header title (in the current spec) isn't achievable without 'stacking' these spaces (e.g. padding 4 is too small, padding 5 is too large). This previously had a bit of a smell.

I think it makes sense to simplify this like what is done here, just wanted to call out that this deviates from the spec slightly!

robinpyon (approved):
Thanks @mariuslundgard! Looks great, just left a few questions below

PR #1182, opened 2023-11-13
fix: update tooltip position middleware, close tooltips on click

robinpyon:
### Description

(This PR extends #1171)

1. Simplify `<Tooltip>` middleware and fix cases where non-portalled tooltips with excess content wouldn't be correctly rendered. This follows floating-ui's official [tooltip example](https://floating-ui.com/docs/tooltip) and no longer uses the `size` middleware. 
2. Ensures that tooltips are automatically closed whenever `onClick` or `onContextMenu` is triggered on child elements.
3. Ensures that user defined event callbacks on tooltip child elements (`onClick`, `onMouseEnter` etc) are correctly triggered.

### What to test

- Non-portalled tooltips with excess content should render with the correct width
- Portalled tooltips should be correctly positioned and ignore any defined boundary element
- Clicking or right-clicking on any tooltip child element should hide any existing tooltip, or prevent one from appearing if it's yet to be made visible (as the result of a custom defined `delay` value)
- Clicking tooltip child elements should still continue to focus them as normal

### Other context

The idea of automatically closing tooltips on click was previously raised in #1108 – though we've since seen some subtle issues arise as a result of consumers not correctly handling this manually (EDX-751). 

I believe this is something that the `<Tooltip>` component in particular should just handle for you automatically – and users can always turn to `<Popover>` components if they need fine-grained control over visibility. Would love any opinions here!

### DIFF
--- a/src/primitives/tooltip/tooltip.test.tsx
+++ b/src/primitives/tooltip/tooltip.test.tsx
@@ -373,7 +373,7 @@ describe('Tooltip', () => {
 
       act(() => jest.advanceTimersByTime(delay))
 
-      // Assertion: the tooltip is no visible
+      // Assertion: the tooltip is not visible
       expect(screen.queryByText('Tooltip content')).toBeVisible()
 
       act(() => fireEvent.click(button))
@@ -399,7 +399,7 @@ describe('Tooltip', () => {
 
       act(() => jest.advanceTimersByTime(delay))
 
-      // Assertion: the tooltip is no visible
+      // Assertion: the tooltip is not visible
       expect(screen.queryByText('Tooltip content')).toBeVisible()
 
       act(() => fireEvent.contextMenu(button))
--- a/src/primitives/tooltip/tooltip.test.tsx
+++ b/src/primitives/tooltip/tooltip.test.tsx
@@ -6,6 +6,17 @@ import {render} from '../../../test'
 import {Text, Button} from '../../primitives'
 import {Tooltip, TooltipDelayGroupProvider} from '../tooltip'
 
+// Fake timers using Jest
+beforeEach(() => {
+  jest.useFakeTimers()
+})
+
+// Running all pending timers and switching to real timers using Jest
+afterEach(() => {
+  jest.runOnlyPendingTimers()
+  jest.useRealTimers()
+})
+
 describe('Tooltip', () => {
   describe('Using same delay for open and close', () => {
     it('should hide and show the tooltip content when hovered, with no delay', () => {
@@ -275,6 +286,7 @@ describe('Tooltip', () => {
       jest.clearAllMocks()
     })
   })
+
   describe('Closing the <Tooltip /> with the Escape key', () => {
     it('Standalone tooltip closes immediately with Escape key', () => {
       const delay = 150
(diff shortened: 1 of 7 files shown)


src/primitives/tooltip/tooltip.tsx
```diff
@@ -63,6 +62,9 @@ export interface TooltipProps extends Omit<LayerProps, 'as'> {
 
 const Root = styled(Layer)`
   pointer-events: none;
+  // Note that 100vw doesn't exclude system scrollbars.
+  // This should be sufficiently small enough to not trigger overflow-x scrollbars when portalled.
+  max-width: calc(100vw - ${DEFAULT_TOOLTIP_PADDING * 10}px);
```
mariuslundgard, line 67: Note that portals are usually smaller than `100vw` in the Studio, since they are often scoped to panes.

mariuslundgard: Could consider measuring the portal and passing a `$portalWidth` property to the `Root` here.

src/primitives/tooltip/tooltip.tsx
```diff
@@ -102,43 +104,30 @@ export const Tooltip = forwardRef(function Tooltip(
     // Flip the floating element when leaving the boundary box
     ret.push(
       flip({
-        boundary: boundaryElement || undefined,
+        boundary: portal ? undefined : boundaryElement || undefined,
```
mariuslundgard, line 107: Hm, I'm not sure what's going on here. Ideally portal and boundary should be separated concerns.

src/primitives/tooltip/tooltip.tsx
```diff
@@ -102,43 +104,30 @@ export const Tooltip = forwardRef(function Tooltip(
...
-        },
-      }),
-    )
-
-    // Shift the tooltip so its sits with the boundary eleement
+    // Shift the tooltip so its sits with the boundary element
     ret.push(
       shift({
-        boundary: boundaryElement || undefined,
+        boundary: portal ? undefined : boundaryElement || undefined,
```
mariuslundgard, line 120: Same as above.

src/primitives/tooltip/tooltip.test.tsx
```diff
@@ -342,4 +354,117 @@ describe('Tooltip', () => {
...
+
+      const button = screen.getByText('Hover me')
+
+      // Assertion: tooltip does not exist in the document
+      expect(screen.queryByText('Tooltip content')).not.toBeInTheDocument()
+      fireEvent.focus(button)
+
+      act(() => jest.advanceTimersByTime(delay))
+
+      // Assertion: the tooltip is no visible
```
pedrobonamin, line 376: tiny thing, should be ?
```suggestion
      // Assertion: the tooltip is now visible
```

pedrobonamin (approved):

robinpyon:
Thank you both!

@mariuslundgard one erroneous assumption I made about this was that portals and boundary elements are mutually exclusive. As you've outlined, this isn't the case – and it's now possible to provide both.

I've updated this PR accordingly – if you provide both a custom named portal AND a boundary element, the tooltip will always strive to use the minimum available width between the two. 

As suggested, it now reads the width of the current portal and passes that to `<Root>`, and a supporting tooltip workshop story (`customPortal`) has been created to highlight usage with both custom portals and boundary elements.

mariuslundgard (approved):

PR #1183, opened 2023-11-14
feat: update avatar styles

robinpyon:
### Description

This PR updates `<Avatar>`:
- adds support for a `hideInnerStroke` prop, which conditionally hides the outline between avatar images and the outer border
- tweaks avatar sizes: reducing diameter, reduces stroke width and tightens the focus ring
- uses medium weight for `<AvatarCounter>` (and also reduces stroke width)

Storybook updates:
- New stories for `<AvatarStack>` and `<AvatarCounter>`

### DIFF
--- a/stories/constants.ts
+++ b/stories/constants.ts
@@ -9,6 +9,9 @@ import {
   Radius,
 } from '../src/types'
 
+export const AVATAR_SRC =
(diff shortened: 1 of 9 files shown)


stories/constants.ts
```diff
@@ -9,6 +9,9 @@ import {
   Radius,
 } from '../src/types'
 
+export const AVATAR_SRC =
```
pedrobonamin, line 12: 😄

pedrobonamin (approved):

PR #1185, opened 2023-11-16
fix(tooltips): update default fallback placements

pedrobonamin:
Update the fallback placements.
It's strange that if we want a tooltip to be positioned e.g. at the bottom, it automatically jumps to the top as a fallback. 
With this changes, it will always try to fallback to the same axis options and then change to other options.

So bottom -> bottom-end, bottom-start Then changes to top.


robinpyon (approved):
Great suggestion – thanks @pedrobonamin!

PR #1187, opened 2023-11-21
feat(animations): add animations to dialog and tooltips.

pedrobonamin:
## Adds animation to Dialog and Tooltips.

- Tooltips include a presence animation, it means, an animation handled by framer-motion will be triggered on mount and unmount of the element.
- Dialogs only include a css enter animation, it was not possible to easily include framer motion to handle the animation as the extra wrappers break the dialog style. Also the `<AnimatePresence>` wrapper would need to be added in the consumer and not in the ui component, given that the Dialog don't receive a `show` prop, and that logic is handled in the consumers. 

https://github.com/sanity-io/ui/assets/46196328/b784c482-b033-4cfc-bca3-dbcf5debc2c6

https://github.com/sanity-io/ui/assets/46196328/8f64a1b9-14e0-4e24-988b-b6d39dcd9eb2

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -54,7 +54,7 @@ export interface DialogProps extends ResponsivePaddingProps, ResponsiveWidthProp
    * Whether the dialog should animate in on mount.
    *
    * @beta
-   * @default false
+   * @defaultValue false
    */
   animate?: boolean
 }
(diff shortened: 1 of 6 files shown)


robinpyon (approved):
Thanks @pedrobonamin! 

Tooltip animations look and feel great, would be awesome if we could facilitate exit animations in dialogs too, but understand how this could be a little tricky

PR #1188, opened 2023-11-28
fix(dialog): drag events in modals are applied to underlying elements

pedrobonamin:
Contain onDragEnter event to the modal only, as it's propagating to the elements behind it causing unexpected behaviours in the elements.

## With error

https://github.com/sanity-io/ui/assets/46196328/c2661532-44de-4fd5-bef9-a693e34eb224

## Fixed:

https://github.com/sanity-io/ui/assets/46196328/c6b317b1-cd5e-42a8-997b-93263ede1c06

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -84,6 +84,10 @@ function isTargetWithinScope(
   )
 }
 
+function onDragEnter(event: React.DragEvent<HTMLDivElement>) {
+  return event.stopPropagation()
+}
+
 const Root = styled(Layer)<ResponsiveDialogPositionStyleProps & ResponsivePaddingStyleProps>(
   responsivePaddingStyle,
   dialogStyle,
@@ -402,6 +406,7 @@ export const Dialog = forwardRef(function Dialog(
         id={id}
         onActivate={onActivate}
         onClick={handleRootClick}
+        onDragEnter={onDragEnter}
         onFocus={handleFocus}
         ref={ref}
         role="dialog"


src/components/dialog/dialog.tsx
```diff
@@ -406,6 +406,7 @@ export const Dialog = forwardRef(function Dialog(
         ref={ref}
         role="dialog"
         zOffset={zOffset}
+        onDragEnter={(event) => event.stopPropagation()}
```
hermanwikner, line 409: Since this is a static function, I suggest moving it outside the component scope so that the function is created only once and won't be recreated on every render.

```ts
const onDragEnter = (event: React.DragEvent<HTMLDivElement>) => event.stopPropagation()

export const Dialog = forwardRef(function Dialog(){
  //  ...
})
```

Nitpick: I think it's always nice to alphabetize the props :)

hermanwikner, line 409: This is working great! However, I'm unsure if there might be a situation where you'd want `onDragEnter` to propagate?

Alternatively, we could consider adding this directly to the `Dialog` component here: 

https://github.com/sanity-io/sanity/blob/b305a8f63430a52ffdf870c9646b2ecf573bbcc2/packages/sanity/src/core/form/components/EditPortal.tsx#L49

pedrobonamin: It's a good point, is worth considering it, I can't think of any case in which we want the dragEvent to propagate from the dialog root.
Moving it to the `EditPortal` I'm not sure if it's possible, as I need this to apply to the root of this element, I mean we could hack a way around it, if you have a good idea in how to implement the same from the `EditPortal` happy to move the change there 
Thanks!

pedrobonamin: Found a way to move it to where you pointed. 
Created PR in mono repo, https://github.com/sanity-io/sanity/pull/5282 thanks @hermanwikner closing this one.

hermanwikner: > Moving it to the EditPortal I'm not sure if it's possible, as I need this to apply to the root of this element

I tested this and added the same `onDragEnter` function to the `Dialog` component in `EditPortal`, and it appears to resolve the issue. The function will be added to the root element in the `Dialog` component (i.e., `Root`) as it receives `{...restProps}` which includes the `onDragStart` function from the props, so I don't think we have to do any hack to make it work. But I guess it's more of a question of whether we should add this in Sanity UI or in the Studio monorepo :)

pedrobonamin:
Closing in favour of PR in mono repo to keep changes contained.
https://github.com/sanity-io/sanity/pull/5282

PR #1194, opened 2023-12-13
feat: add animate support to popovers, add custom origin middleware and animate from reference origin

robinpyon:
### Description

This PR adds `animate` support for `Popover` components. 

In addition to fading in, animated popovers will now also slightly scale from the reference element's origin, giving the appearance of popovers very slightly 'growing out'. Since this is a popover prop, it's also possible to apply this animation on `<MenuButton>` components too (via the `popover` prop) too.

This is done via a custom `origin` middleware which calculates this transform origin offset from both the shifted position and final (not user supplied) floating placement.

This middleware is also applied to `<Tooltip>` components too.

Currently, there's no optionality around animation here – opting in means you get both opacity + scale transitions, though we may want to provide more fine-grained control in future.

https://github.com/sanity-io/ui/assets/209129/a4ad64ae-ca78-429a-8d0d-00335f8f6fbf

### DIFF
--- a/src/core/primitives/popover/popoverCard.tsx
+++ b/src/core/primitives/popover/popoverCard.tsx
@@ -1,10 +1,12 @@
 import {Strategy} from '@floating-ui/react-dom'
 import {ThemeColorSchemeKey} from '@sanity/ui/theme'
+import {MotionProps, motion} from 'framer-motion'
 import React, {CSSProperties, forwardRef, memo, useMemo} from 'react'
 import styled from 'styled-components'
+import {POPOVER_MOTION_PROPS} from '../../constants'
 import {BoxOverflow, CardTone, Placement, PopoverMargins, Radius} from '../../types'
 import {Arrow, useLayer} from '../../utils'
-import {Card} from '../card'
+import {Card, CardProps} from '../card'
 import {Flex} from '../flex'
 import {
   DEFAULT_POPOVER_ARROW_HEIGHT,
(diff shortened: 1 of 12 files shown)


src/core/primitives/tooltip/tooltipCard.tsx
robinpyon, line 1: Separated this out into its own component, to match `popoverCard.tsx`

src/core/primitives/popover/popoverCard.tsx
```diff
@@ -105,9 +115,9 @@ export const PopoverCard = memo(
     )
 
     return (
-      <Root
+      <MotionCard
```
robinpyon, line 118: Note that we're not conditionally creating this motion component – we're always leveraging `motion(Card)` even if you don't opt into `animate`. 

Given that we're not rendering these motion elements as wrappers, it felt simpler to just conditionally attach motion props as needed rather than conditionally define this component. WDYT?

mariuslundgard: I think it's fine – there might be a tiny perf hit, but nothing compared to perf issues already caused by styled-components

src/core/primitives/tooltip/tooltipCard.tsx
mariuslundgard: 👍

mariuslundgard (approved):
beautiful – thank you for this!

PR #1200, opened 2024-01-03
fix: improve scale transitions for non-retina screens

stipsan:
Fixes scale transitions that struggle with text aliasing, this is particularly visible if your monitor doesn't have a retina resolution:

https://github.com/sanity-io/ui/assets/81981/64a9a353-4000-4366-848b-a97c0832f7ba

The above shows the problem for popovers, but it's also affecting tooltips and toasts.

After the fix in this PR is applied:

https://github.com/sanity-io/ui/assets/81981/d97e4224-0751-4225-b5ef-641d3f8db903


robinpyon (approved):
Great catch, thanks @stipsan!

PR #1206, opened 2024-01-15
fix(theme): restore selectable in v0 theme

pedrobonamin:
fixes: https://github.com/sanity-io/sanity-plugin-markdown/issues/90

Restores Legacy `sanity.color.selectable `values  in the v1 theme object when using the conversion function `v2_v0` which takes the `V2 theme` and creates the `V0 theme`.


robinpyon (approved):
Thanks @pedrobonamin!

PR #1217, opened 2024-01-18
fix(popover): default open popover has incorrect position

pedrobonamin:
An issue was detected in arcade, in which the popover won't position correctly when it's rendering as open on mount.

This is caused by a `useEffect` that is removing the floating ui reference set by the child when it's mounted. 
Checking if child exists fixes the issue.

This only happens in SSR environments.

## Pre fix
Screenshot 2024-01-18 at 10 49 30

## Post fix
Screenshot 2024-01-18 at 10 51 22


robinpyon (approved):
Thanks @pedrobonamin! 

It'd be great to add a clarifying comment to `referenceElement` (in `PopoverProps`) that it's a noop if `children` are defined

PR #1244, opened 2024-03-26
fix(primitives): disabled button contrast 

hermanwikner:
This pull request improves the contrast of disabled buttons. Note that this pull request aims to _improve_ the contrast ratio; however, the contrast ratios might not comply with a11y standards. Another pass needs to be done in the future to ensure the buttons have the correct ratio in regard to a11y standards.

Screenshot 2024-04-04 at 12 13 23


ninaandal (approved):
It does improve it - but as you say, it should probably be reconsidered, because it still has a low contrast imo..

PR #1245, opened 2024-03-26
fix: use default `styled-components` export to restore `styled-components@5` compatibility

juice49:
- `@sanity/ui` [allows](https://github.com/sanity-io/ui/blob/main/package.json#L195) `styled-components@^5.2` as a peer dependency.
- In [commit ](https://github.com/sanity-io/ui/commit/1ad5a02320434104dbe15dbd519e2db488fb15e7)1ad5a02[ of ](https://github.com/sanity-io/ui/commit/1ad5a02320434104dbe15dbd519e2db488fb15e7)`@sanity/ui`, we switched to the named `styled` export, which is [only available in ](https://styled-components.com/releases#v6.0.0-beta.7)`styled-components@6`.
- Therefore, some folks have `styled-components@^5.2` installed in their Studio project, but the installed version of `@sanity/ui` isn't compatible. It can’t import `styled`.

In this branch, I've reverted to the default export, which is compatible with both `styled-components@5` and `styled-components@6`.

This can be tested in alpha release `2.0.13-alpha.2`. For testing purposes, you may need to set an override in your Studio project. For example, using pnpm:

```
"pnpm": {
  "overrides": {
    "@sanity/ui": "2.0.13-alpha.2"
  }
}
```


binoy14:
Does it also make sense to change the devDependency to `styled-components@v5` so if someone by mistake uses v6 API it fails early?

binoy14 (approved):
Left one non-blocking comment but the rest of the changes makes sense to me thanks!

juice49:
@binoy14 Good point! It makes sense to develop using the lowest common denominator of Styled Components.

juice49:
Not releasing this for now, as there's concern about breaking the package for [ESM consumers](https://github.com/styled-components/styled-components/issues/3437).

> In transpiled/commonJS environments that works as expected, **but breaks server-side ESM usage of styled components**.

stipsan (changes requested):
We should drop support for styled-components v5 instead of doing this. Otherwise we lose support for embedding `sanity` in Remix + Vite apps and others.
Or implement a workaround such as https://github.com/styled-components/xstyled/blob/d213d97f0caee4249cddb33a696dc47439f59057/packages/emotion/src/emStyled.ts#L2

PR #1288, opened 2024-05-10
fix(menu): remove use of `isElement`

mariuslundgard:
Remove unnecessary use of `isElement` in the `MenuButton` component.


christianhg (approved):
I agree. This looks unused so we might as well remove it.

PR #1323, opened 2024-05-29
feat(card): add `muted` property

mariuslundgard:
With this change we introduce a new `muted` property on the `<Card>` component.

`<Card muted tone="inherit">` will render a card with a slightly darker background (and slightly lighter in dark) than its surroundings.

The benefit of this in comparison to `<Card tone="transparent">` is that you can combine `muted` and `tone` like `<Card muted tone="critical">` – and it also works within button, card, and menu item states.

Try it here:
https://sanity-ui-workshop-git-feat-muted-card.sanity.build/primitives/card/props


ninaandal (approved):
lgmt ✨

PR #1324, opened 2024-05-31
feat: deprecate `useForwardedRef`

stipsan:
The reasons why this hook is a bad idea is well explained in this thread:
https://x.com/buildsghost/status/1792946027127046249

In short it's a bad idea to create `useRef` hooks inside custom hooks as it makes the Compiler unable to optimize the code when the custom hook lives in a different file.

[Notice how in this example, it's fully optimized (as indicated by the presence of `react.memo_cache_sentinel` in the compiled output)](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwBUYCASggGYA0JZAolVQnIQL4FUwQYEDkMBAEM2-ADoA7TDnxFSCAMIAbNAkl4A6mgAmeABaduvAQDoA9ApVqN2vfomSpCAB6zCO6sKjLCVKJJsaBCSBIq8OJLqeAAUWDxYYHRUuADuwjCeOpRUAJREUgQEcCFghEJUBAC89BTUMbmFxaWEcKrRdgbVtdYdugYxFY1OobVMLGwxDdUAfAWjRSWSkMoIpsoQAOaD1KZwsEIaw0XsdADaFQC6w01CeLChADw6aABuBBVVwBWc5jNS7CkIHYQA).
[While the lack of local instantiation of `useRef` causes it to be less optimized (no `react.memo_cache_sentinel`)](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwBUYCAogGaUJx4A0BluA7gIYwAmASgpY6QQBJbAhhs8aAG4IAEmwB2nADYICAXyYwIGAgHIYCNnT0AdBZhz4iggMLK0CBXgDqaTngAWmytt16AOgB6OwcnV3cvMwVzBAAPK0JOPjYoZUJKKAU6NAgFAlsdHAVwgAosbSwwRmYYdi4EHj4ASiJzAgI4PLBCAG1Dfk6w5zcPTwBdAgBeEjJ7RxHIz1Lm83bZimpaPFKV6YA+NvyOzu6IVQDlCABzUoGAuFhDZ1Xj9UZ+vnHX9cM8WHyAB5ONICAMpsABpogvtzOpzCB1EA)

So by using `useRef` and `useImperativeHandle` directly, instead of `useForwardedRef` we solve almost all cases where the React Compiler opts-out (the only one remaining is `useArrayProp`). Also worth mentioning that by using `useImperativeHandle` instead of `useIsomorphicEffect`, which uses `useLayoutEffect` under the hood, we also enable codebases like `sanity`, that still uses `useForwardedRef`, to no longer have the downside of the double side-effects execution that `useLayoutEffect` causes, and reduces jank.


mariuslundgard (approved):
This is great – thanks for doing this!

PR #1346, opened 2024-06-11
feat: add loading bar to toast

ninaandal:
Screenshot 2024-06-11 at 10 12 16

Adding a loading bar to the toast card to indicate when the toast will close. 

I need some input from you css experts here, because ideally the loading bar should be the border itself in case the `Card` has a different radius. I've hardcoded this radius: 
```
border-bottom-left-radius: var(--card-border-radius, 6px);
border-bottom-right-radius: var(--card-border-radius, 6px);
``` 
I know this might not be the ideal solution, so I appreciate feedback!


src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+`
+
+const loadingAnimation = keyframes`
+  0% {
+    background-position: 100% 0;
+  }
+  100% {
+    background-position: 0 0;
+  } 
+`
```
robinpyon, line 18:
```suggestion
const loadingAnimation = keyframes`
  0% {
    width: 0;
  }
  100% {
    width: 100%;
  } 
`
```

I can't speak for performance but I think this is probably easier to reason about – we're not animating background images so it's probably enough!

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export function rootStyles(
+  props: {$duration?: number; tone: string} & ThemeProps,
```
robinpyon, line 21:
```suggestion
  props: {$duration?: number; tone: ThemeColorStateToneKey} & ThemeProps,
```

This should get around the type issues below!

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+    width: 100%;
+    position: relative;
+    padding-bottom: 0px;
+    padding-left: 0.5px;
+    &::before {
+      content: '';
+      position: absolute;
+      bottom: 0px;
+      border-bottom-left-radius: var(--card-border-radius, 6px);
+      border-bottom-right-radius: var(--card-border-radius, 6px);
```
robinpyon, line 47: As Cody mentioned above, no need to set the border radius here – this can be removed and you can instead lean on adding `overflow` to the parent element

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+      width: 100%;
+      height: 3px;
+      background-image: linear-gradient(
+        to right,
+        ${loadingBarColor} 50%,
+        var(--loading-bar-color-to, transparent) 50%
+      );
```
robinpyon, line 54: This could probably be simplified to `background: ${loadingBarColor}` as from what I can tell, `--loading-bar-color-to` hasn't been defined

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+      background-position: 100% 0;
+      background-size: 200% 100%;
```
robinpyon, line 56: This can also be removed if you simplify `loadingAnimation` and instead animate on `width` only (again – not 100% sure on perf here! but probably not an issue)

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+      animation-name: ${loadingAnimation};
+      background-attachment: fixed;
+      animation-timing-function: ease-in-out;
+      animation-iteration-count: infinite;
```
robinpyon, line 60: nit: can be omitted since the animation should in theory never repeat (the toast will be gone) – the default value is `1` in this case

robinpyon (commented):
Looking great Nina!

Just left a bunch of nits, feel free to do with it as you will!

A few other points:

1. The text doesn't appear _optically_ centered in a single line toast, might be worth seeing what design has to say, but pretty sure we'd want to add some padding to offset this

image

2. We may want to consider scenarios when multiple toasts are triggered with the same ID – in these cases, the internal `setTimeout` will reset but the transition wont. You can see what I mean by going [here](https://sanity-ui-storybook-git-edx-694.sanity.build/?path=/story/components-toast--with-hook) and clicking any toast multiple times

3. An awesome fast follow would be to allow users to cancel the automatic closing on `duration` whenever the popover is clicked, which I think is pretty standard behaviour

ninaandal:
Thank you both @stipsan @robinpyon, this was great feedback! Looks so much better now 🙇 

Rob, about your points: 
1. Added some bottom padding, looks better now I think! 
2. Yes - I agree. I'll add a Linear issue for this (it wasn't introduced in this pr) 
3. That sounds right! I'll ping Kayla about this 😄

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,58 @@
...
+    width: 100%;
+    position: relative;
+    overflow: hidden;
+    overflow: clip;
+    padding-bottom: 3px;
+    &::before {
+      content: '';
+      position: absolute;
+      bottom: 0px;
+      height: 3px;
```
robinpyon, line 47: nit: given that `padding-bottom` and `height` have to be the same, creating a separate const would make it more ergonomic to change this in future, and better signify the intent behind specifying a custom `padding-bottom`

robinpyon: Also, could we live with a 2px height? It's minor, but the bar does seem disproportionately large. Will defer to design here!

src/core/components/toast/styles.ts
```diff
@@ -0,0 +1,58 @@
...
+      background: ${loadingBarColor};
+      animation-name: ${loadingAnimation};
+      animation-duration: ${props.$duration}ms;
```
robinpyon, line 50: suggestion: If we're not re-starting the animation whenever a toast with the same `id` is triggered, one way to possibly patch over this is by adding either `animation-fill-mode: both` or `animation-fill-mode: forward` ([source](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode)) to ensure that the progress bar remains at 100% on completion, even if a user spams that multiple times

Example:

https://github.com/sanity-io/ui/assets/209129/2cd42e6b-3d45-43cd-b69a-91fa1bd97546

robinpyon (dismissed):
Thanks @ninaandal – please remove the errant `package-lock.json` but otherwise looks great!

Left a few minor (non-blocking) comments too

robinpyon (approved):

PR #1349, opened 2024-06-13
fix: react 19 compat issues with children prop

stipsan:
While investigating a bug where the image input menu buttons:
image

just refuses to render on React 19 ([repro](https://test-next-studio.sanity.build/test/structure/author;c04836eb-ddba-423a-bcc6-4759b02c55c1)) I discovered that the way we loop over the `children` prop in `<Inline />` have always been unsafe. After fixing it the menu buttons started showing again ([repro](https://test-next-studio-git-fix-missing-image-input-buttons-in-82d3fa.sanity.build/test/structure/author;0294295e-93c8-4612-b3b1-47fb08912d7d)).

While at it I fixed similar issues in other components 🙌


src/core/components/tree/tree.tsx
```diff
@@ -213,7 +213,7 @@ export const Tree = memo(
       ) as HTMLElement[]
 
       setItemElements(_itemElements)
-    }, [children, ref])
```
stipsan, line 216: fixing this while at it, I missed it when refactoring forwarded refs

src/core/primitives/avatar/avatarStack.tsx
```diff
@@ -66,9 +65,10 @@ export const AvatarStack = forwardRef(function AvatarStack(
     size: sizeProp = 1,
     ...restProps
   } = props
-  const children = childrenToElementArray(childrenProp).filter(
-    (child) => typeof child !== 'string',
-  ) as React.ReactElement[]
+  const children = useMemo<React.ReactElement[]>(
+    () => Children.toArray(childrenProp).filter(isValidElement),
```
stipsan, line 69: The `isValidElement` check already filters out strings, and also filters out `0`, `false`, `null` and `undefined`

rdunk (approved):
LGTM

src/core/primitives/inline/inline.tsx
```diff
@@ -28,10 +27,7 @@ export const Inline = forwardRef(function Inline(
   const {as, children: childrenProp, space, ...restProps} = props
 
   const children = useMemo(
-    () =>
-      childrenToElementArray(childrenProp)
-        .filter(Boolean)
-        .map((child, idx) => <div key={idx}>{child}</div>),
+    () => Children.map(childrenProp, (child) => child && <div>{child}</div>),
```
stipsan, line 30: When using `Children.map` it's not necessary to set `key`, as the original `key` on the child element is merged with your key. IOW it's only necessary to set `key` if you're returning a fragment:
```tsx
Children.map(childrenProp, (child, idx) => child && <><div key={`${idx}-child`}>{child}</div><hr key={`${idx}-separator`} /></>)
```

src/core/components/tree/tree.tsx
rdunk: 👍

src/core/primitives/avatar/avatarStack.tsx
rdunk: Sounds good.

src/core/primitives/inline/inline.tsx
stipsan: > Even when children is an array, Children.map has useful special behavior. For example, Children.map combines the [keys](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key) on the returned elements with the keys on the children you’ve passed to it. This ensures the original JSX children don’t “lose” keys even if they get wrapped like in the example above.

https://arc.net/l/quote/khsvrxsp

rdunk: Cool, noted here: https://react.dev/reference/react/Children#why-is-the-children-prop-not-always-an-array

PR #1353, opened 2024-06-25
feat(textarea): support for disabled focus ring

jordanl17:
Sheet List feature in Studio currently uses the `__unstable_disableFocusRing` prop on `TextInput` to turn off the input styling. This is useful because the Sheet List view cell provides all focus, selection and range styling whilst the input itself is used only for value changing.

Now functionality is being extended to support `TextArea` and as such the same prop would be very useful, but does not currently exist. This PR adds in the `__unstable_disableFocusRing` props for `TextArea` which can be used in the same way as for `TextInput`


mariuslundgard (approved):
LGTM!

PR #1372, opened 2024-07-11
fix: revert recent portal/tooltip changes

rexxars:
The changes introduced in v2.3.4, v2.3.5 and v2.3.6 caused issues for the studio, where tooltips inside of modals would not appear anymore. To unblock users, I'd rather revert these changes now and investigate how we can reduce layout thrashing in a more structured manner once released.


ricokahler (approved):
assuming this is a simple revert, let's merge!

PR #1373, opened 2024-07-12
fix(tooltip): use layout effect to calculate width

stipsan:
Trying again! Turns out it was the change to `portalProvider` that caused the regression in `sanity` where tooltips stopped rendering inside modals (which is also why its e2e tests were failing on the renovate PR).

[It's not clear why this change caused it to break](https://github.com/sanity-io/ui/pull/1369/files#diff-34317e40981e35c81d7a84cdf83737f5a6c04ad0848608faf362f8ab8db00658):
```diff
- const fallbackElement = useSyncExternalStore(
-   emptySubscribe,
-   () => document.body,
-   () => null,
- )
+ const mounted = useMounted()

  const value: PortalContextValue = useMemo(() => {
    return {
      version: 0.0,
      boundaryElement: boundaryElement || null,
-     element: element || fallbackElement,
+     element: element || mounted ? document.body : null,
      elements,
    }
- }, [boundaryElement, element, elements, fallbackElement])
+ }, [boundaryElement, element, elements, mounted])

  return <PortalContext.Provider value={value}>{children}</PortalContext.Provider>
```

I'll revisit this in a follow up and get to the bottom of it, and while at it set up more tests that can catch these regressions here so that the studio doesn't have to serve as the canary in the coal mine 😅 

The changes in this PR has already [been tested by the studio testing suite](https://github.com/sanity-io/sanity/actions/runs/9903569379/job/27359371667?pr=7147) out of an abundance of caution: https://github.com/sanity-io/sanity/pull/7147

As well as manually [verified](https://test-studio-git-test-sanity-ui-canary-tooltip-fix.sanity.build/test/structure/author;jordan-drop-a.428510f1-860b-4b16-8d53-6c4c0fc12acf) that tooltips still works inside modals:

https://github.com/user-attachments/assets/801feda4-fb10-4177-8d87-2924590f29a4


rdunk (commented):
This LGTM from a code perspective but without much context (but so did the other PRs!). Also tested as you described and the behaviour seems correct. I'm working on the assumption that the portalProvider change was an unrelated layout thrashing mitigation, which is why it's safe to merge this without.

I'm also assuming the single failing test on the linked PR is unrelated to this, but unsure as to why the whole suite passes even though there are failures, is that expected?

LMK if you want someone with more knowledge to review before merging, will approve if not.

stipsan:
@rdunk good catch! Will re-run that testing suite locally first and get to the bottom of that. Disabled auto-merge as well for now 🙌

.eslintrc.js
```diff
@@ -137,6 +137,6 @@ module.exports = {
         mode: 'full',
       },
     ],
-    react: {version: 'detect'},
+    'react': {version: 'detect'},
```
snorrees, line 140: Wait, the new default is REQUIRING quotes on properties? Thats cray 🙊

snorrees (commented):
Why not do the prettier change in a separate PR? I dunno whats going on with the default, but this new default triggers me.

stipsan:
> Why not do the prettier change in a separate PR? I dunno whats going on with the default, but this new default triggers me.

It's an artifact of how https://github.com/sanity-io/ui/pull/1372 reverted 3 separate PRs, where the first one had the prettier change 😅 

Looks like the failing component test succeeded when retried, we're good to go!

.eslintrc.js
stipsan: No, it's [set to `consistent`](https://prettier.io/docs/en/options.html#quote-props) by the request of @mariuslundgard :)

src/core/primitives/tooltip/tooltip.tsx
```diff
@@ -454,4 +430,47 @@ export const Tooltip = forwardRef(function Tooltip(
...
+  showTooltip,
+}: {
+  handleIsOpenChange: (open: boolean, immediate?: boolean) => void
+  referenceElement: HTMLElement | null
+  showTooltip: boolean
+}) {
+  // Since we don't want the `mouseevent` events to be attached and removed if the `referenceElement` is changed
+  // we use a "effect event" (https://19.react.dev/learn/separating-events-from-effects#reading-latest-props-and-state-with-effect-events)
+  // in order to always see the latest `referenceElement` value inside the event handler itself.
+  const onMouseMove = useEffectEvent((target: EventTarget | null, teardown: () => void) => {
```
snorrees, line 449: Ah, so `useEffectEvent` is the "I want to do a sideffect when effect deps change, but I dont want the variables used in the sideeffect itself to trigger it)?

src/core/primitives/tooltip/tooltip.tsx
```diff
@@ -402,9 +376,11 @@ export const Tooltip = forwardRef(function Tooltip(
       data-ui="Tooltip"
       {...restProps}
       ref={setFloating}
-      style={floatingStyles}
+      style={{
+        ...floatingStyles,
+        maxWidth: tooltipMaxWidth > 0 ? `${tooltipMaxWidth}px` : undefined,
```
snorrees, line 381: Why was it necessary to move this out of the styled css?

snorrees (approved):

src/core/primitives/tooltip/tooltip.tsx
rdunk: That's how I understand it. Basically non-reactive `useEffect` triggered event handlers.

src/core/primitives/tooltip/tooltip.tsx
stipsan: Because it's ridiculously expensive to update the CSSOM every time the `maxWidth` changes, with CSS-in-JS, it's far more performant to just set it directly on `style` :)

TL;DR `styled-components` should create as static CSS strings as possible, to reduce the amount of CSSOM writes as possible.
[The official react docs talks about it more](https://react.dev/reference/react/useInsertionEffect#injecting-dynamic-styles-from-css-in-js-libraries):
> If you use CSS-in-JS, we recommend a combination of the first two approaches (CSS files for static styles, inline styles for dynamic styles). **We don’t recommend runtime `<style>` tag injection for two reasons**:

> 1. Runtime injection forces the browser to recalculate the styles a lot more often.
> 2. Runtime injection can be very slow if it happens at the wrong time in the React lifecycle.

By setting it on `style` we avoid both these two problems that they outline ☝️

snorrees: Lol, we have some work to do in Create. I've always been a proponent of style directly, but Sanity code reviews end up moving them to styled components 🤷

PR #1377, opened 2024-07-15
chore: setup sanity e2e tests

stipsan:
Testing if it's [possible to reuse](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow) the E2E workflows in the sanity monorepo to protect against regressions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,21 +13,35 @@ permissions:
   contents: read # for checkout
 
 jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -62,6 +75,15 @@ jobs:
             cypress/screenshots
             cypress/videos
 
+  sanity-e2e:
+    if: ${{ github.event_name == 'pull_request' }}
+    permissions:
+      contents: read
+      actions: write
+    name: "Run Studio End-to-end tests"
+    uses: sanity-io/sanity/.github/workflows/e2e-ui.yml@next
```
bjoerge, line 84: TIL you can do this ✨

bjoerge (approved):

PR #1385, opened 2024-07-16
feat: add `useMatchMedia` hook

stipsan:
Split from #1368. 

The new `useMatchMedia` hook is also now underpinning `usePrefersDark` and `usePrefersReducedMotion`. It uses less memory than the previous implementations as module scoped cache is no longer used:
```ts
let MEDIA_QUERY_CACHE: MediaQueryList | undefined
```
Browsers does a really good job at optimizing multiple `window.matchMedia` instances that use the same media query, so there isn't really a reason for us to cache it. It's better to let React manage it and allow the browser to garbage collect `window.matchMedia` instances that no longer is subscribed to.

Lastly, by moving `usePrefersDark` and `usePrefersReducedMotion` they now allow customizing the `getServerSnapshot` implementation. Here's an example of how it can be used to detect Dark Mode server side, and ensure SSR hydration doesn't have a "flash of light mode" before dark mode is detected in the browser: https://gist.github.com/stipsan/13c0cccf8dfc34f4b44bb1b984baf7df


src/core/hooks/useMatchMedia.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Efficiently subscribes to `window.matchMedia` queries
+ *
+ * @param getServerSnapshot - Only called during server-side rendering, and hydration if using hydrateRoot. Required if the hook is called during SSR (https://react.dev/reference/react/useSyncExternalStore#adding-support-for-server-rendering)
+ *
+ * @public
+ */
+export function useMatchMedia(
+  mediaQueryString: `(${string})`,
```
snorrees, line 11: Whats the reason for not adding these `()`inside the function automatically?

src/core/hooks/useMatchMedia.ts
```diff
@@ -0,0 +1,46 @@
...
+     * `subscribe` and `getSnapshot` are only called on the client and both need access to the same `matchMedia` instance
+     * we don't want to eagerly instantiate it to ensure it's only created when actually used
+     */
+    let MEDIA_QUERY_CACHE: MediaQueryList | undefined
+
+    const getMatchMedia = (): MediaQueryList => {
+      if (!MEDIA_QUERY_CACHE) {
+        // As this function is only called during `subscribe` and `getSnapshot`, we can assume that the
+        // the `window` global is available and we're in a browser environment
+        MEDIA_QUERY_CACHE = window.matchMedia(mediaQueryString)
```
snorrees, line 25: TIL matchMedia 👍

snorrees (approved):
veri nais!

src/core/hooks/useMatchMedia.ts
stipsan: Because it's nice when the APIs are the same:
```
window.matchMedia("(max-width: 600px)")
useMatchMedia("(max-width: 600px)")
```

PR #1389, opened 2024-07-17
chore: add `useClickOutside` test

stipsan:
Before sending a PR for updates to the implementation of `useClickOutside` based on #1368 it makes sense to first add a comprehensive testing suite that documents how its current API surface works, and its flaws.


snorrees (approved):
Very nice, some of these has definetly caught me by surprise before (ie, req to use state and not refs ect).

PR #1391, opened 2024-07-17
fix: restore `useClickOutside` behaviour, move new logic to `useClickOutsideEvent`

stipsan:
Turns out the changes to `useClickOutside` in v2.8.0 had regressions that weren't detected in our end to end testing suites, but discovered by @binoy14 during manual testing: https://github.com/sanity-io/sanity/pull/7186#pullrequestreview-2183886727

This PR restores `useClickOutside` to how it was implemented in v2.7.0 and before. While the improvements based on `use-effect-event` is moved to a new, separate, hook called `useClickOutsideEvent`.


binoy14 (approved):
this is great, thanks!

PR #1401, opened 2024-07-18
fix: add `displayName` to all components to improve debugging

stipsan:
Just like https://github.com/sanity-io/icons/pull/92, it's really difficult to make sense of component trees in the React DevTools when function names are minified:

image

After display names are manually added life gets much easier, regardless of how the app that uses `@sanity/ui` is bundling its code for production:

image


snorrees (approved):
⚡ 🚤

PR #1432, opened 2024-10-29
fix(primitives): button text ellipsis

hermanwikner:
### Description

This pull request fixes an issue where text does not truncate with an ellipsis in the button primitive. The issue appears to have been introduced in #1195 when the element wrapping text was removed.

text-ellipsis


robinpyon (approved):
LGTM! Looks like there's existing precedent for this approach [here](https://github.com/sanity-io/ui/blob/fix/button-text-ellipsis/src/core/components/menu/menuGroup.tsx#L187-L191).

PR #1440, opened 2024-10-31
fix(Dialog): start alignment on dialog header

jordanl17:
Currently with a long title, the Dialog uses `center` which means the close button moves down from the top.
Screenshot 2024-10-30 at 00 29 07

Now, these elements are `flex-start` so the close button is always the same padding from the top of the dialog


mariuslundgard (approved):
Thanks for fixing 🙏

PR #1478, opened 2024-11-28
fix(popover): render the child when a reference element is provided

mariuslundgard:
This fixes a bug whiched caused the `Popover` component to not render the child element when a `referenceElement` property was provided.


stipsan (approved):
Good catch!

PR #1479, opened 2024-12-04
fix(theme): "upgrade" v2 colors to v2.9 format

rexxars:
This might be throwaway code, but you get the gist - in some contexts with multiple versions on the same page, one theme does not have the new `neutral` and `suggest` props and should be upgraded. See slack for more detail. I am not sure if this _actually_ solves it yet as I have to run, but would be interested in hearing your thoughts


mariuslundgard (approved):
Ah, this is spot on. Thank you!

stipsan (approved):
Agree with Marius, just saw a `TypeError: Cannot read properties of undefined (reading 'v2')` error in the wild, so I'll merge and release this 🙌

PR #1480, opened 2024-12-04
fix(card): guard when picking new nested theme properties

mariuslundgard:
After introducing the new tones (`neutral` and `suggest`) we need to treat them as potentially missing since it's possible to use multiple versions of `@sanity/ui` at the same time. This is necessary when using a new `Card` component within an older `ThemeProvider`.


bjoerge (commented):
LGTM, thank you @mariuslundgard!

bjoerge (approved):

PR #1560, opened 2025-01-10
fix: remove `legacyExports`

stipsan:
We no longer support older bundlers like webpack v3, or older jest test runners.


mariuslundgard (approved):
SGTM!

PR #1603, opened 2025-02-13
feat(button): add `textWeight` property

mariuslundgard:
Add `textWeight` property to buttons it's possible to adjust font weight of button labels:

```tsx
<Button text="Button" textWeight="regular" />
```


colepeters (commented):
Love to see it! It looks like the new `textWeight` prop is missing from the workshop, though? E.g. https://sanity-ui-workshop-git-feat-button-text-weight-property.sanity.dev/primitives/button/props?scheme=light

mariuslundgard:
Ah, good point @colepeters – will add that!

colepeters (approved):

PR #1616, opened 2025-02-17
fix: use linear timing for toast duration bars

stipsan:
Changes the timing of the duration bar on toasts so it's linear, and thus predictable, instead of the default timing (which is `animation-timing-function: ease` since we're not defining one).

The `ease` timing slows down both the start of the animation, and the end, quite a bit. It leads to the effect that it's difficult to guess when it's about to close, and how long it'll stay open.

## Before:

Note how it looks like all 4 notifications will close about the same time, even though it's not

https://github.com/user-attachments/assets/2d869210-28e0-4f5f-bca0-06ea27db8f5c

## After:

While now it's clear when they'll close and exactly how long they've been open.

https://github.com/user-attachments/assets/fd5ef4d6-b242-4437-81c8-0bcb29dc54bf

### DIFF
--- a/src/core/components/toast/styles.ts
+++ b/src/core/components/toast/styles.ts
@@ -51,6 +51,9 @@ export function rootStyles(
       animation-name: ${loadingAnimation};
       animation-duration: ${props.$duration}ms;
       animation-fill-mode: both;
+      animation-timing-function: linear;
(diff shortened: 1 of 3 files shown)


robinpyon:
Thanks @stipsan! Honestly, I prefer the version _with_ easing, but it's not something I'm that strongly opinionated about. Will defer to @mariuslundgard here

mariuslundgard (approved):
LGTM

PR #1618, opened 2025-02-19
feat: optional support for pointer-blocking popovers

robinpyon:
https://github.com/user-attachments/assets/e00d9bd8-f344-4bbf-b389-80cc3fdad634

## Description

This PR adds support for rendering popovers that prevents pointer interaction with all outside elements until closed.

Popovers rendered with `popover={{modal: true}}` will now include a hidden, full viewport overlay that blocks all pointer events. Clicking anywhere in the viewport will then close the popover as normal.

This can be especially useful in interfaces where popovers appear over the top of other interactive elements (such as links), making it much more forgiving for users to close popovers without fear of unwanted side effects. This is the default behaviour for contextual menus in OS X and a number of other popular apps.

As this is opt-in, existing popovers and components that use them (`<MenuButton>`, `<Autocomplete>`) are unaffected.

## What to test

- A [story](https://sanity-ui-storybook-git-feat-blocking-popovers.sanity.dev/?path=/story/components-menubutton--popover-modal) has been created with an example scenario. Compare behaviour between the default menu button and those with blocked pointer events. Does it feel intuitive to you?
- ~~`blockPointerEvents` may not be the best name for this, open to suggestions!~~ Renamed to `modal` at Marius' request

### DIFF
--- a/src/core/primitives/popover/popover.tsx
+++ b/src/core/primitives/popover/popover.tsx
@@ -58,13 +58,6 @@ export interface PopoverProps
    */
   animate?: boolean
   arrow?: boolean
-  /**
-   * When true, blocks all pointer interaction with elements beneath the popover until closed.
-   *
-   * @beta
-   * @defaultValue false
-   */
-  blockPointerEvents?: boolean
   /** @deprecated Use `floatingBoundary` and/or `referenceBoundary` instead */
   boundaryElement?: HTMLElement | null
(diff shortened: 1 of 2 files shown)


colepeters (approved):
Code & results LGTM! Tested with pointer, touchscreen, and keyboard inputs and works as I'd expect.

PR #1619, opened 2025-02-21
fix: use stricter type definition for children of <Code>

bjoerge:
Currently, the `<Code>` component accepts any valid `ReactNode`, which means this is works compile time:
```tsx
<Code><b>Ooops</b></Code>
```

But breaks runtime, as the Code component expects `children` to be stringifiable, coerces it to a string runtime and thus render:
```
[object Object]
```

This lead to error messages in the studio showing `[object Object]` instead of the error message, see for example the bug report over at https://github.com/sanity-io/sanity/issues/8712

This fixes the issue by changing the `children` type to something that can safely be stringified, and also considering that `children` can be iterable (e.g. `<Code>The number is {42}</Code>` will make `children` iterable).

Note: _incidentally_, the above snipped actually used to work prior to lazy loading refractor (783942ab82e732a4e258e0d35dff44d0e5d16ca1), because before then, we'd only stringify if refractor was loaded and `language` was passed: https://github.com/sanity-io/ui/blob/7a0231bdc05cc7005c7c1c2c2cda89e9e38a0730/src/primitives/code/code.tsx#L24-L26

This caused inconsistent behavior, so we should stick with always stringifying children.

After this PR, the above code will give a compile error:
```
error TS2322: Type 'Element' is not assignable to type 'StringifiableNode'.

       <b>Ooops</b>
        ~~~~~~~~~~~~

  code.tsx:27:3
      children: StringifiableNode
         ~~~~~~~~
    The expected type comes from property 'children' which is declared here on type 'IntrinsicAttributes & Omit<CodeProps & Omit<HTMLProps<HTMLElement>, "as" | "size" | "children">, "ref"> & RefAttributes<...>'
```

### DIFF
--- a/src/core/primitives/code/__workshop__/opticalAlignment.tsx
+++ b/src/core/primitives/code/__workshop__/opticalAlignment.tsx
@@ -1,4 +1,3 @@
-import {AddCircleIcon} from '@sanity/icons'
 import {Box, Card, Code, Flex, Stack} from '@sanity/ui'
 
 export default function OpticalAlignment() {
@@ -37,9 +36,7 @@ export default function OpticalAlignment() {
 
         <Flex>
           <Card padding={2} scheme="dark">
-            <Code>
-              <AddCircleIcon />
-            </Code>
+            <Code>{`<AddCircleIcon />`}</Code>
           </Card>
         </Flex>
       </Stack>
--- a/src/core/primitives/code/code.tsx
+++ b/src/core/primitives/code/code.tsx
@@ -1,4 +1,4 @@
-import {forwardRef, Suspense, lazy} from 'react'
+import {forwardRef, lazy, Suspense} from 'react'
 import {styled} from 'styled-components'
 import {useArrayProp} from '../../hooks'
 import {responsiveCodeFontStyle, ResponsiveFontStyleProps} from '../../styles/internal'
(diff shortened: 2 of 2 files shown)


bjoerge:
Thinking some more about this, I'm wondering if we, for now, should instead restore the previous behavior of not stringifying if `language` is omitted. The inconsistency really feels like a bug to me, but I suspect there are quite a few cases of this breaking in the wild. WDYT @stipsan?

stipsan:
@bjoerge it seems like the stringify is an oversight:
```tsx
<LazyRefractor language={language} value={String(children)} />
```
The previous behavior should be restored by just passing children as is?
```tsx
<LazyRefractor language={language} value={children} />
```

bjoerge:
Yeah, that should restore the earlier behavior

EoinFalconer (dismissed):
Much better testing. I don't know if we want to add tests for the stringifyChildren function?

bjoerge:
Now that #1620 landed this is less of a pressing issue, but still think we should consider this improvement as it removes an inconsistency leading to confusing behavior.

I start out by writing this:
```tsx
<Code>
  <b>Some code</b>
</Code>
```
This outputs **Some code** in bold, perhaps as expected.

However, if I then add a language:
```tsx
<Code language="typescript">
  <b>Some code</b>
</Code>
```

This will now render `[object Object]` and not give me any hits about what's wrong. Even worse, this is also dependent on whether the selected language is registered with refractor or not, e.g. `language="nosuchlang"` will revert back to the jsx-rendered output.

To me, this feels like very much like a bug not a feature, and I think the right thing is to always stringify children regardless. I don't see the case for using this component to render anything else than source code, but maybe I'm missing something.

Note: this would be a breaking change, so updated the commit to reflect that.

stipsan:
I'm hesitant to do a v3 of `@sanity/ui` for this, as it's a large library and it'll take time for plugins and even our own tooling to coalesce on it. In the mean time it'll double bundle sizes, which in the case of `sanity schema extract` commands, bundle commands like `next build`, significantly add to these.
I'm more inclined to have it rollout in a minor, with a phase where we warn against non-string code, and use the new typings, first. And then rollout the breaking change in a minor.

bjoerge:
That's fair. I guess you can also argue that it's a bugfix :) I'll adjust the implementation and commit message accordingly (I'm giving this lower priority, so probably won't happen before next week some time).

PR #1633, opened 2025-02-28
feat: improve toast, popover, and tooltip motion

stipsan:
Updates the Motion transitions we have for Toasts, Popovers and Tooltips to be more fluid and responsive. Removes `opacity: -1` and `opacity: 2` hacks.

They're very smooth in slow motion:

https://github.com/user-attachments/assets/b6219c83-7a13-4e97-8d5c-539a31a7a20e

# Tooltip

Left side is before, right side after. The animation no longer have this "flick" or "flash" effect as the text fades in.

https://github.com/user-attachments/assets/420f9b64-cd92-4543-83e8-e2e36a604e21

# Popover

Fade in is slowed down here as well, while the overall animation is sped up.

https://github.com/user-attachments/assets/9085c1d4-9a15-40d7-aeba-ebde83e06f0d

# Toasts

The `AnimatePresence` now uses `mode="popLayout", which animates sibling toasts right away instead of waiting for it to fade out:

https://github.com/user-attachments/assets/2bd7da0e-766e-46f5-9342-f6c2ec25c90a

It works really well when there's multiple toasts of diverse colors and various duration timing set:

https://github.com/user-attachments/assets/6772a562-2005-4f03-ab05-843b7d8e4b82

When a toast is mutated (using a shared `id`) the progress bar resets to reflect the actual time left before the toast auto dismisses:

https://github.com/user-attachments/assets/6ba53556-1afc-46a2-8215-bf20c0251037

It's now possible to use `toast.push({duration: Infinite})` to make a toast stay open until the user dismisses it:

 

https://github.com/user-attachments/assets/ab40d85e-471a-4187-b945-52158a364fb1

Due to a lack of support for `duration: Infinite`, the Sanity Studio codebase has long used workarounds where the duration is set to the highest possible number that setTimeout allows, this pattern is still supported to ensure we don't break BC:

https://github.com/user-attachments/assets/bcd87663-fd03-4f1f-b36f-e0ed38f56cae

`@sanity/ui` doesn't yet support programmatically closing a toast by `id`. The studio codebase gets around this by setting a really short duration of `0.01`. This pattern is still supported to ensure full backwards compatibility:

https://github.com/user-attachments/assets/c7af23b3-9e24-4a61-a0d5-6964ef1bcc22

### DIFF
--- a/src/core/components/toast/styles.ts
+++ b/src/core/components/toast/styles.ts
@@ -1,64 +1,76 @@
-import {styled, keyframes, css} from 'styled-components'
+import {styled} from 'styled-components'
 import {ThemeColorStateToneKey, getTheme_v2} from '../../../theme'
-import {POPOVER_MOTION_CONTENT_OPACITY_PROPERTY} from '../../constants'
-import {Flex} from '../../primitives'
-import {ThemeProps} from '../../styles'
+import {Card, Flex} from '../../primitives'
+import type {ButtonTone} from '../../types'
+
+const LOADING_BAR_HEIGHT = 2
+
+export const STATUS_CARD_TONE = {
+  error: 'critical',
+  warning: 'caution',
+  success: 'positive',
+  info: 'neutral',
+} satisfies {[key: string]: ThemeColorStateToneKey}
+
+export const BUTTON_TONE = {
+  error: 'critical',
+  warning: 'caution',
+  success: 'positive',
+  info: 'neutral',
+} satisfies {[key: string]: ButtonTone}
 
 export const TextBox = styled(Flex)`
   overflow-x: auto;
 `
 
-const loadingAnimation = keyframes`
-  0% {
-    width: 0;
-  }
-  100% {
-    width: 100%;
+export const StyledToast = styled(Card)`
+  pointer-events: all;
+  width: 100%;
+  position: relative;
+  overflow: hidden;
+  overflow: clip;
+
+  &[data-has-duration] {
+    padding-bottom: calc(${LOADING_BAR_HEIGHT}px / 2);
   }
 `
 
-const LOADING_BAR_HEIGHT = 2
+export const LoadingBar = styled.div`
+  display: flex;
+  position: absolute;
+  bottom: 0px;
+  top: 0px;
+  left: 0px;
+  right: 0px;
+  pointer-events: none;
+  z-index: -1;
+  overflow: hidden;
+  overflow: clip;
+  background: transparent;
+  align-items: flex-end;
+  will-change: opacity;
+`
 
-export function rootStyles(
-  props: {$duration?: number; tone: ThemeColorStateToneKey} & ThemeProps,
-): ReturnType<typeof css> {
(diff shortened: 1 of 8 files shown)


src/core/components/toast/styles.ts
```diff
@@ -1,64 +1,76 @@
...
-    width: 0;
-  }
-  100% {
-    width: 100%;
+export const StyledToast = styled(Card)`
+  pointer-events: all;
+  width: 100%;
+  position: relative;
+  overflow: hidden;
+  overflow: clip;
```
colepeters, line 31: Just curious about the use of both `hidden` and `clip`. Do we have some clients running browsers that don't support `clip`?

src/core/components/toast/toastProvider.tsx
```diff
@@ -1,195 +1,107 @@
...
-
+      startTransition(() => {
         setState((prevState): ToastState => {
-          const idx = prevState.findIndex((t) => t.id === id)
+          const dismiss = () => {
+            startTransition(() => {
+              setState((prevState): ToastState => {
+                const idx = prevState.findIndex((t) => t.id === id)
+
+                if (idx > -1) {
```
colepeters, line 46: Would love to see some extra commenting on the control flow here just for clarity. I got it after walking through it a couple times but a comment might speed up groking the flow here ;)

src/core/components/toast/toast.tsx
```diff
@@ -52,25 +46,63 @@ const StyledToast = styled(Card)<{$duration?: number; tone: ThemeColorStateToneK
...
+    title,
+    status,
+    updatedAt,
+    ...restProps
+  } = props
   const cardTone = status ? STATUS_CARD_TONE[status] : 'default'
   const buttonTone = status ? BUTTON_TONE[status] : 'default'
   const role = status ? ROLES[status] : 'status'
 
+  const prefersReducedMotion = usePrefersReducedMotion()
```
colepeters, line 77: 1 million bonus points for using `prefers-reduced-motion`!!

src/core/components/toast/toastProvider.tsx
```diff
@@ -1,195 +1,107 @@
...
 
-            return toasts
+          // Backwards compatibility for `sanity` patterns workaround a lack of programatically dismissible toasts
+          if (duration === 0.01) {
+            return prevState.filter((t) => t.id !== id)
           }
 
           return prevState
+            .filter((t) => t.id !== id)
+            .concat([
```
colepeters, line 66: Possibly some commenting here would help clarify the intention of the operations as well!

colepeters (dismissed):
This looks great! Thanks so much for the detailed description and videos, they were so helpful in understanding all the changes. Also such a treat to see these visual upgrades, the animations are so well considered and executed. I'm approving but have some non blocking questions and requests for extra comments. Happy to file another approval if needed if you end up making any updates!

colepeters (approved):
Beautiful! Thanks so much for the updates and extra comments, helps a ton!

PR #1674, opened 2025-03-26
chore: eslint config

mariuslundgard:
This PR upgrades the eslint configuration to work with `eslint@9`. It also adds eslint configuration to enforce module boundary rules (useful for keeping module exports cleanly separated and avoid circular deps).

Most interesting commit to review: https://github.com/sanity-io/ui/pull/1674/commits/5764902e113b8565fb1e949f2d1e5aa404f0109d


stipsan (approved):
Let's go!

PR #1683, opened 2025-03-31
fix: consistent delayed tooltip group behavior

stipsan:
There's a race condition in `<Tooltip />`, unclear when it snuck into the codebase, but it's causing only every second or so tooltip to actually open if you move the cursor in a group quickly enough:

https://github.com/user-attachments/assets/4cc87045-8cbd-4ba3-8eb7-da343c899177

Here it is with the fix:

https://github.com/user-attachments/assets/a610916c-8352-4bca-a34a-219c5b79d98b

We should come back and fix the state management and cleanup `<Tooltip />` in general, it's really complex. For now this fix should do.

I removed the `isGroupActive` state from the context provider since it's not actually used anywhere.


pedrobonamin (approved):
This looks good to me, I agree in doing a cleanup of tooltip, it looks more complicated than what it could be

PR #1711, opened 2025-04-30
fix(deps): upgrade to react compiler RC

stipsan:
[The compiler is now in RC](https://react.dev/blog/2025/04/21/react-compiler-rc), time to hop off the beta train 🥳 

The `eslint-plugin-react-hooks` case is due to us using `use-effect-event`: https://github.com/sanity-io/use-effect-event?tab=readme-ov-file#usage-with-eslint-plugin-react-hooks, otherwise we'd use the `rc` tag there as well :)


src/core/hooks/useMatchMedia.ts
stipsan, line 1: This file is refactored to make it pass the React Compiler linter test. The old version has an error, the new version passes.

As it turns out, due to the subscribe step:
```
        const media = window.matchMedia(mediaQueryString)
        media.addEventListener('change', onStoreChange)
        return () => media.removeEventListener('change', onStoreChange)
```

There isn't a perf cost to construct a media query on `getSnapshot`:
```
() => window.matchMedia(mediaQueryString).matches
```

We're not constantly constructing and throwing away observers, as the browser handles instances and caching for us.
IOW our userland caching is unnecessary 😮‍💨

mrmrs (approved):
This looks great @stipsan - thanks for being so thorough!

PR #1741, opened 2025-06-13
feat: allow popover to use auto placement strategy

mogenshoj:
https://github.com/user-attachments/assets/c5aeac8f-edee-4961-8553-7ba18914ee57

## Description
This PR adds the ability to choose positioning strategy for the `Popover` component.

It can either use `flip` (the currently used and the default after this PR), or the user can optionally pick `autoPlacement`. Which uses the [autoPlacement](https://floating-ui.com/docs/autoPlacement) middleware.

`autoPlacement` can be useful in interfaces where the popover content is dynamic (eg. filtering) and there is a desire for the first picked placement to be retained while the user filters the content.

**Example**
When using `flip` placement strategy with these props
```
fallbackPlacements={['top', 'top-end', 'bottom-start', 'bottom', 'bottom-end']}
placement="top-start"
```

And the initial content cannot be placed at the `top`, `flip` will automatically place it at `bottom`.
If the user is filtering the content by typing, once it becomes small enough to fit `top`, the placement will shift while the user is still typing.

Using `autoPlacement`, the position will be based on where the most available viewport space is, and it won't matter what the height of the content is. So the originally picked placement will be retained while the user types (or until the user scrolls the viewport enough, so that the most available viewport space change).

## Testing
A story named `Auto Placement` has been created for the `Popover` component which shows the behavior.

### DIFF
--- a/src/core/primitives/popover/popover.tsx
+++ b/src/core/primitives/popover/popover.tsx
@@ -7,6 +7,7 @@ import {
   offset,
   RootBoundary,
   shift,
+  autoPlacement,
   useFloating,
 } from '@floating-ui/react-dom'
 import {ThemeColorSchemeKey} from '@sanity/ui/theme'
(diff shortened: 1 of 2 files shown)


colepeters (dismissed):
Changes LGTM, thanks Mads! This approval should also get Vercel to build the preview deployment.

Edit: apparently not? 😅 Not sure what the build issue is here!

robinpyon (approved):
Thanks @mogenshoj!

mariuslundgard (approved):
Looks good! Thank you 🙇

stipsan (approved):

stipsan:
Gonna override and merge, there's something wrong with the CI setup and it has to be fixed in the monorepo, as the e2e-ui.yml is defined there 🙌

PR #1767, opened 2025-07-09
fix(popover): pass tone to PopoverLayer

joshuaellis:
### Description

We used to pass `tone` to PopoverLayer and we still should because otherwise the tone is never passed correctly and everything is stuck with `initial`.


hermanwikner (approved):
Nice 🙇‍♂️

PR #1941, opened 2025-08-13
feat(menuItem): allow the ability to prevent default behaviour

joshuaellis:
### Description

_Sometimes_ you don't want your menu to close. We can easily achieve this by calling `event.preventDefault` on the `onClick` handler MenuItem accepts and calls _before_ `onItemClick` and then check in the latter callback thus preventing the menu closing.

### Related Issues

* DASH-459

### What to review

Does it make more sense to check `defaultPrevented` in the `handleClick` function of `MenuITem`?

### Testing

I did add a test, but popover's don't render in the test environment because of their `referenceHidden` middleware, maybe something we want to look at if we want more unit tests?


stipsan (approved):
Looks good! We need a better test setup anyway 😮‍💨

PR #2055, opened 2025-09-04
fix: remove `useArrayProp` usage

stipsan:
The React Compiler is able to conditionally memoize props that call `getArrayProp`, while `useArrayProp` has to follow the rules of hooks.


mariuslundgard (approved):
LGTM!

PR #2061, opened 2025-09-10
feat: add options prop to useGlobalKeyDown

RitaDias:
### Description

Add options as part of the props that you can receive to the useGlobalKeyDown hook. This is particularly important in the studio where we can have multiple GlobalKeyDowns and it's important to add options such as capture: true

This is currently needed to fix a bug in the studio [Sanity PR 10606](https://github.com/sanity-io/sanity/pull/10606)

### What to review

The code, should I add anything, is there any code missing?

### Testing

N/A


bjoerge (approved):
Nice, lgtm!

PR #2140, opened 2025-10-17
fix(toast): clear existing timeout id when pushing toast

mariuslundgard:
### Description

Pushing a toast with an ID that already exists, adds a duplicate setTimeout, and unwanted results.

This fix addresses this by clearing existing timeout id when pushing a new toast.


stipsan (commented):
Looks good! Although I'd prefer a `Map` to keep track of the timeout ids instead of `{}` 🙂

PR #2169, opened 2025-11-06
fix(button): revert adding explicity inline-size

gu-stav:
### Description

The `inlineSize` prop was added in https://github.com/sanity-io/ui/commit/e3ba3cb4d1958a84aa6edbdffeb7dfef9e9287a0, but breaks a ton of buttons in Dashboard, because adding `width="fill"` to buttons no longer works.

Refs https://github.com/sanity-io/ada/pull/1965

https://sanity-io.slack.com/archives/C08SH30FJQL/p1761558755847009

### DIFF
--- a/src/css/primitives/button/button.css.ts
+++ b/src/css/primitives/button/button.css.ts
@@ -23,7 +23,6 @@ export const root: string = _style(layers.primitives, {
   verticalAlign: 'top',
   backgroundColor: vars.color.bg,
   color: vars.color.fg,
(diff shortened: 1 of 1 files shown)


joshuaellis (approved):
There's no description on the commit that helps me understand why the change was originally made, but given it breaks dashboard and v4 should have no visual breaking changes lets go

mariuslundgard (approved):
Yup, this is good to revert 👍

PR #2171, opened 2025-12-05
chore: update storybook

kristofferlund:
### Description
Updated to the latest version of Storybook and updated the Vercel deploy commands to serve v4

### What to review
See that it works at all.
I replaced some non-existing variables of tones and colors, but I was just guessing here so please guide me if I'm wrong. 

### Testing
Did not add or change any testing


stipsan (approved):
Thank you! 💖

PR #2177, opened 2026-02-25
fix: lazily compute color themes, tones, v0 colors

rexxars:
### Description

Currently, `buildColorTheme` takes an average of 236ms to execute on my (M4 Pro) MacBook. A lot of the computation is either not needed or can be deferred until needed. This PR adds lazy evaluation in a few different parts:

1. Non-default color schemes and tones
2. v0 compatibility `color` property

This brings the up-front computation cost down to under 1ms. If we wanted to "prime" schemes/tones so it doesn't cost us at an inopportune render-time, one could do something like:

```js
import {buildTheme, getScopedTheme} from '@sanity/ui/theme'

const theme = buildTheme()

requestIdleCallback(() => {
  getScopedTheme(theme, 'light', 'default')
  getScopedTheme(theme, 'dark', 'default')
})
```

Getting the scoped theme takes about ~16ms for each one.

### What to review

Changes appear safe and don't have unwanted side effects

### Testing

No new tests. This is transparent for the end user and should be considered an implementation detail.

### DIFF
--- a/src/theme/build/buildColorTheme.ts
+++ b/src/theme/build/buildColorTheme.ts
@@ -27,6 +27,7 @@ import {
   ThemeColorSyntax,
 } from '../system'
 import {ColorTokenContext, resolveColorTokenValue as _color} from './colorToken'
+import {defineLazyProperty} from './lib/lazy'
 import {resolveColorTokens} from './resolveColorTokens'
 
 export function buildColorTheme(config?: ThemeConfig): ThemeColorSchemes_v2 {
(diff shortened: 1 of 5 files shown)


rexxars:
The studio E2E workflow seems to be failing because of upstream things - I've attempted to solve them [in this PR](https://github.com/sanity-io/sanity/pull/12240), but not quite sure.

colepeters (approved):
Nice work! Going off [the explanation of `v2`/`v0` you gave me](https://sanity-io.slack.com/archives/C07RS3HV1K9/p1772050221120069?thread_ts=1772048867.316959&cid=C07RS3HV1K9), everything here looks correct and improved to me. Stoked to have some perf upgrades going out for this! 🏆

PR #2179, opened 2026-02-27
V4 beta docs

somerandomdude:
### Description

Adding TypeDoc documentation to v4-beta component props

### What to review

We will need to vet the efficacy of these new typedoc comments.


colepeters:
@somerandomdude I think we can close this, right?

PR #2621, opened 2026-08-05
fix: handle display array in box codemod

laurenashpole:
### Description

This PR updates the Box codemod to check for `flex` and `grid` display values in responsive arrays in addition to strings. It doesn't handle mixed arrays (arrays with `block` and `flex` for example) but there aren't currently any instances of those in studio and they would be caught during typechecking. 

### What to review

I would focus on whether the logic in `box.ts` and the test make sense. The `getStaticAttributeExpression` file contains more agent aided jscodeshift code.

### Testing

N/A

---

> [!NOTE]
> **Low Risk**
> Scoped to ui-codemod transform utilities and tests; no runtime app behavior, though other codemods using `getStaticAttributeExpression` could see array values where they previously got `null`.
> 
> **Overview**
> The **Box** codemod can now treat responsive `display` props written as static arrays (e.g. `['none', undefined, null, 'flex']`) the same as a plain `'flex'` / `'grid'` string when deciding to rewrite to **Flex** or **Grid**.
> 
> **`getStaticAttributeExpression`** is extended to return parsed static array literals (strings/numbers/booleans, `null`, and identifier `undefined`), not only single primitives; spreads or non-static elements still yield `null`.
> 
> Replacement logic in **`box.ts`** is centralized in **`matchesDisplayValue`**, which normalizes scalar vs array values, requires a matching `flex`/`grid` suffix, and **skips** the rewrite when the array mixes incompatible display families (e.g. `block` and `flex` in the same array). New inline tests cover the happy path and the mixed-array guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65652a2c9d8784013f64571accb71cc62aab87d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ui-codemod/src/transforms/latest/box/box.test.ts
+++ b/packages/ui-codemod/src/transforms/latest/box/box.test.ts
@@ -182,6 +182,18 @@ defineInlineTest(
   transform,
   {},
   `
+  <Box display={['none', 'flex']} />
+  `,
+  `
+  <Flex display={['none', 'flex']} />
+  `,
+  'replaces Box with Flex when display is an array with flex',
+)
+
+defineInlineTest(
+  transform,
+  {},
+  `
   import {Box} from '@sanity/ui'
 
   function Example() {
--- a/packages/ui-codemod/src/transforms/latest/box/box.test.ts
+++ b/packages/ui-codemod/src/transforms/latest/box/box.test.ts
@@ -182,10 +182,10 @@ defineInlineTest(
   transform,
   {},
   `
-  <Box display={['none', 'flex']} />
+  <Box display={['none', undefined, null, 'flex']} />
   `,
   `
-  <Flex display={['none', 'flex']} />
+  <Flex display={['none', undefined, null, 'flex']} />
   `,
   'replaces Box with Flex when display is an array with flex',
 )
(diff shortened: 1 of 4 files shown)


packages/ui-codemod/src/transforms/latest/box/box.test.ts
```diff
@@ -178,6 +178,18 @@ defineInlineTest(
...
+  {},
+  `
+  <Box display={['none', undefined, null, 'flex']} />
+  `,
+  `
+  <Flex display={['none', undefined, null, 'flex']} />
+  `,
+  'replaces Box with Flex when display is an array with flex',
+)
+
```
colepeters, line 192: Worth adding a test for the mixed case?

colepeters (approved):
Presuming the jscodeshift code is legit, this seems good to go to me! One question, but not blocking

packages/ui-codemod/src/transforms/latest/box/box.test.ts
laurenashpole: Good call, that caught a bug that I just fixed.

colepeters (approved):
Nice!

colepeters (approved):
Still nice!

PR #2657, opened 2026-08-10
fix: box/flex/grid default min-width and min-height

laurenashpole:
### Description

This PR adds default `min-height: 0` and `min-width: 0` to the Box, Flex, and Grid components to match v3 and v4.

### What to review

Make sure the default props make sense on those components.

### Testing

I checked that the components had `min-height` and `min-width` styling set by default and that I could still override it by setting the props.

---

> [!NOTE]
> **Low Risk**
> Small default-style change on layout primitives; overrides remain possible, with possible minor visual/layout shifts where defaults were relied on implicitly.
> 
> **Overview**
> **`Box`**, **`Flex`**, and **`Grid`** now default **`minHeight`** and **`minWidth`** to **`'0'`**, wired through **`getProps`** so the layout props apply the same way as other sizing defaults. This restores behavior from **v3/v4** and helps nested flex/grid layouts shrink and scroll instead of overflowing when min sizes would otherwise stay at **`auto`**.
> 
> Consumers can still override both props explicitly. The change is released as a **patch** on **`@sanity/ui`** via changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad37974282cc2adc9f269d3aace5a20c690cd0e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/curly-trams-feel.md
+++ b/.changeset/curly-trams-feel.md
@@ -1,0 +1,5 @@
+---
+'@sanity/ui': patch
+---
+
+add default min-height and min-width to box/flex/grid
--- a/packages/ui/src/components/box/Box.tsx
+++ b/packages/ui/src/components/box/Box.tsx
@@ -10,9 +10,14 @@ const boxClassName = suffixClassName('sui-Box')
 /** @public */
 export function Box<T extends ElementType = 'div'>({
   display = 'block',
+  minHeight = '0',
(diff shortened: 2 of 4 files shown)


colepeters (approved):
LGTM!

PR #2658, opened 2026-08-10
fix: v5 alpha codemod flex

laurenashpole:
### Description

This PR fixes a small codemod bug when transforming shorthand attributes into separate props. When getting the mapped value for those props, the helper was returning on a false positive based on the value and not the key. That's kind of confusing but basically, I wanted to transform `flex={1}` into `flexGrow={1}` but I was getting `flexGrow={1} flexShrink={1}` instead because `flexShrink` had mappings with the value `1`.

I also updated the `flexBasis` mappings since `auto` (the default) and `0%` can cause different styling and fixed a couple instances in the tests where I was passing strings (which is how this bug was missed in the first place).

### What to review

This is mostly robot code so make sure the general ideas make sense.

### Testing

The codemod tests should be more accurate now since they reflect the correct types for the `flex` prop.

---

> [!NOTE]
> **Low Risk**
> Scoped to codemod mapping logic and tests; no runtime UI or auth/data paths.
> 
> **Overview**
> Fixes **`@sanity/ui-codemod`** so expanding the **`flex`** shorthand no longer applies extra props when another sub-prop’s mapping merely shares the same **value** (e.g. `flex={1}` incorrectly also setting **`flexShrink`**).
> 
> **`getMappingValue`** now resolves mappings by **key** first (string/number/template literal), and only falls back to value matching for **booleans**. **`flexBasis`** mappings add numeric **`1`–`12`** → **`0%`**. Inline transform tests use numeric **`flex`** values instead of strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd87425134f880021346d97738ad34391f289bed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ui-codemod/src/constants/latest/layout-mods.ts
```diff
@@ -109,6 +109,18 @@ export const LAYOUT_MODS: AttributeMods = {
...
+          3: '0%',
+          4: '0%',
+          5: '0%',
+          6: '0%',
+          7: '0%',
+          8: '0%',
+          9: '0%',
+          10: '0%',
+          11: '0%',
+          12: '0%',
```
colepeters, line 123: I'm a little unclear on these 0% mappings — can you elaborate?

packages/ui-codemod/src/utils/transformAttributes.test.ts
```diff
@@ -314,7 +314,7 @@ defineInlineTest(
   transform,
   {},
   `
-  <div flex={["2", "none"]} />
+  <div flex={[2, "none"]} />
```
colepeters, line 317: Wait, where does the 2 come from in this test?

colepeters (commented):
Couple questions!

packages/ui-codemod/src/utils/transformAttributes.test.ts
laurenashpole: It's a test so it can be any valid flex prop value. But it was previously passing 2 as a string instead of a number.

packages/ui-codemod/src/constants/latest/layout-mods.ts
laurenashpole: These are the values that are assigned to `flex-basis` when the shorthand `flex` is used with any number. I'd assumed it would match the general `flex-basis` default but that's `auto` and in some cases that causes styling differences.

colepeters: Ohhhh okay I get it now, thanks!

packages/ui-codemod/src/utils/transformAttributes.test.ts
colepeters: Oh geez, from the truncated diff view I was looking at the code above it and thinking 'why did flex grow/shrink 1 become 2? Expanded it below and I get it now 🤦🏻

colepeters (approved):
LGTM!

PR #2720, opened 2026-08-17
feat: add Modal component

colepeters:
### Description

Adds a new Modal component. This supersedes the Dialog component in UI 3/4. The change in name is to help distinguish this component as being a _modal_ dialog — that is, a dialog that blocks interaction with the rest of the page from the time it's opened to the time it's closed.

This component makes use of [the HTML dialog element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog). The dialog element comes with several advantages out of the box:

- Rigorously tested for interaction and accessibility by W3C standards bodies and browser implementers — focus trapping, escape key functionality, etc., all comes prebaked
- Able to render both modal and nonmodal dialogs (only modal in the case of this component, however)
- No 3rd party dependencies to track/quarantine/add to JS bundles

The Modal component introduces the following API changes from the previous Dialog component:

- `animated` is no longer a prop — the component always animates in (unless the user prefers reduced motion)
- card radius and shadow options have been removed — the dialog's presentational design should be consistent across all instances
- we recommend using `Modal.Content` to render the main content of the modal rather than the `children` prop; `Modal.Content` places the content in an auto scrolling container that will occupy up to the maximum available space in the Modal before scrolling kicks in
- the `footer` prop has been replaced by the `Modal.Footer` subcomponent
- the `id` prop has been removed — no longer necessary
- the `scheme` prop has been removed — modals should take on the scheme of the surface they’re rendered in
- the `width` prop has been removed — this should be controlled using a Container component in the Modal body if needed
- the `zIndex` prop has been removed — Modals will always render in the topmost stacking layer, ignoring z-index altogether

Modal’s `open` prop is new — it’s «used to indicate whether the modal is open or not by the consumer. As long as a boolean value is provided, it could derive from a state variable, a route, a hash, or any other mechanism suitable for the consumer. Thus, rendering the Modal component does not require a `{ open && <Modal … /> }` short circuit the way the previous Dialog component did.

The dialog element can be dismissed by the user by clicking on its close button, by hitting the Escape key, or by clicking on its backdrop (except, for now, in Safari — [support for `closedby` landed in Safari Technology Preview in April](https://webkit.org/blog/17934/release-notes-for-safari-technology-preview-242/), but the public build is still waiting for this to land). Thus, as with the previous Dialog component, the consumer needs to pass a function to the Modal’s `onClose` prop to set the value of `isOpen` back to false.

**NB:** For now, if a `header` is not provided, we still render the header row of the Modal because it contains the Modal’s close button. This can result in odd presentations (as well as reduced accessibility aids). This carries forward the existing behaviour of Dialog in UI 3, but we'll revisit this in a fast follow next week.

Further PR(s) will follow with codemods for this component.

### What to review

- Any missing functionality to reach parity w/ UI 3/4?
- Anything unclear from the implementation?

### Testing

- Opted into basic accessibility and performance testing in Storybook
- Added to DOM depth and INP test harnesses — very good results!

---

> [!NOTE]
> **Medium Risk**
> Adds a new overlay primitive that controls focus trapping, document inertness, and close behavior via the native dialog element. Bugs here would affect accessibility and interaction blocking.
> 
> **Overview**
> Adds a **beta `Modal`** that wraps the native `<dialog>` element (`showModal` / `close`) as the UI 5 successor to UI 3’s `Dialog`. Consumers keep the component mounted and drive visibility with `open` plus `onClose`; dismiss happens via close button, Escape, or backdrop (`closedby="any"`).
> 
> The API is narrower than Dialog: always-on enter animation, fixed radius/shadow, no `id` / `scheme` / `width` / `zIndex`. Body and actions go in `Modal.Content` and `Modal.Footer` instead of a `footer` prop. Header remains optional but the close-button row still renders.
> 
> Storybook covers basic open/close (asserting `:modal`), scrolling content, and footer. Depth and INP harnesses get matching Dialog vs Modal demos.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d440cc0de7598017aec05f7a53fa4d2d0c9c9f0a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ui/src/components/modal/modal.props.ts
```diff
@@ -0,0 +1,20 @@
+import {type PropDef} from '../../types/PropDef'
+
+/** @public */
+export interface ModalProps extends Omit<React.ComponentProps<'dialog'>, 'open'> {
```
laurenashpole, line 4: If `open` is the default dialog prop, would it make more sense to use that than `isOpen` since it'll be expected? That would also be in line with how ui v4 has handled things like [Popovers](https://www.sanity.io/ui/docs/primitive/popover) in the past.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const modalRef = useRef<HTMLDialogElement>(null)
+
+  const modalClasses = clsx(modalClassName, 'sui-inset0 sui-m-auto sui-radius5 sui-shadow3')
+
+  useEffect(() => {
+    const modalElement = modalRef.current
+    if (!modalElement) return
+
+    if (isOpen) {
+      if (!modalElement.open) {
```
laurenashpole: I'm not sure if we need the `modalElement.open` checks in this `useEffect`. It doesn't seem like it should be possible for them to get out of sync with `isOpen`.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+        modalElement.showModal()
+      }
+    } else {
+      if (modalElement.open) {
+        modalElement.close()
+      }
+    }
```
laurenashpole, line 52: You may want a cleanup function to make sure the modal is closed if it unmounts unexpectedly.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+      ref={modalRef}
+      closedby="any"
+      onClose={onClose}
+      className={clsx(modalClasses, className)}
+      style={style}
+      data-ui="Modal"
+      {...rest}
+    >
+      <Flex justifyContent="space-between" alignItems="center" padding={4}>
+        <Heading trim size={1}>
```
laurenashpole: A few accessibility questions around the heading. Is it okay always rendering as an `h2`? If it's set, should the dialog point to it with `aria-labelledby`? And, should this be conditionally rendered to avoid an empty `h2` in the tree?

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }, [isOpen])
+
+  return (
+    <dialog
+      ref={modalRef}
+      closedby="any"
+      onClose={onClose}
+      className={clsx(modalClasses, className)}
+      style={style}
+      data-ui="Modal"
```
laurenashpole, line 70: I feel like the required/recommended attributes change all the time but would we want an `aria-modal` here? Was looking at the [W3 dialog example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/#rps_label).

laurenashpole (commented):
Looking good! Left a few comments/questions.

One last question I had, I see we're intentionally removing the `footer` prop but it is used about six times in our internal apps so I wonder if we'll need to support it?

packages/ui/src/components/modal/modal.props.ts
colepeters: The issue here is that `open` is an HTML attribute and [it's recommended that it not be set directly — instead, the show/close methods should be used](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/open). So that's why I'm omitting it from the props and using an `isOpen` instead which doesn't feed into but rather acts as a signal for the methods.

packages/ui/src/components/modal/Modal.tsx
colepeters: `isOpen` can be true but `modalElement.open` can still be false because there's no symmetrical handling between them except via this effect. So `isOpen` can change independently of the `open` dialog property — either from a change in the `isOpen` state variable by the consumer, or via the user dismissing the modal via the close button, Esc key, or clicking the background (which itself only fires the modal's `close()` method but doesn't directly interact with the `isOpen` state variable). Does that make sense?

packages/ui/src/components/modal/Modal.tsx
colepeters: Great questions!

1. Dialogs are sectioning roots so any `h` level is okay. I went with `h2` because I tend to reserve `h1` for page titles, but it could be argued that it could be an `h1` or any other `h` level.
2. Yes! Thank you!
3. `header` is a required prop so I think not?

packages/ui/src/components/modal/modal.props.ts
laurenashpole: It wouldn't ever be used to set directly since it would be pulled out of the `restProps` and used the same way as `isOpen`. It's not too big a deal, I've just noticed that in Studio there's seems to be preference for `open` over `isOpen` whenever they're dealing with state so it would be nice to stay consistent with what we've done in the past.

packages/ui/src/components/modal/Modal.tsx
colepeters: Good question! I went to double check.

From MDN:
> <dialog> elements invoked by the showModal() method implicitly have aria-modal="true"

packages/ui/src/components/modal/Modal.tsx
laurenashpole: I'm not sure I'd agree with `header` being required. It isn't in v4 and, looking at modals on the site, there seem to be a lot of uses where we wouldn't want it to be.

packages/ui/src/components/modal/Modal.tsx
colepeters: Great call!

packages/ui/src/components/modal/modal.props.ts
colepeters: Ahh, I see what you mean. I wasn't sure if it would be confusing for external devs though since developers might already be used to not passing `open` as it's not meant to be set directly (even if that's not what we'd be doing). If we or others end up finding `open` makes for better ergonomics I'm fine with that! :)

packages/ui/src/components/modal/Modal.tsx
colepeters: Ah good spot. I'll make it optional and conditionally return it then ☺️  Thanks!

packages/ui/src/components/modal/Modal.tsx
laurenashpole: Nice! Love having less things to keep track of.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+    } else {
+      if (modalElement.open) {
+        modalElement.close()
+      }
+    }
+
+    return () => {
+      modalElement.close()
+    }
```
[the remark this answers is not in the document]
colepeters: Dear robot,
That's intentional because we want the modal to fire `close()` if it's unmounted.
Sincerely,
Cole

packages/ui/src/components/modal/modal.css
```diff
@@ -0,0 +1,45 @@
...
+.sui-Modal:not([open]) {
+  opacity: 0;
+  scale: 0.9;
+}
+
+.sui-Modal[open] {
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-4);
+  overflow-y: auto;
```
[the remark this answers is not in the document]
colepeters: Dear robot,
You might be right if a consumer doesn't use Modal.Content, but that's up to them to handle as we want to leave the door open for custom behaviour. The expected pattern is to use Modal.Content, which sets `flexGrow=1` and has `overflow-y: auto` which will scroll content if it exceeds the space available in the parent (Modal.Root) (since the Modal.Root has max width/height in `dvw`s. Since Modal.Content will scroll if its content is maximally occupying available space in Modal.Root, Modal.Root will never overflow and thus need to scroll itself.
Sincerely,
Cole

apps/depth-testing/src/routes/ui3.tsx
```diff
@@ -26,6 +30,22 @@ function Ui3() {
         <Box marginBottom={4}>
           <Text as="p">A measure of the USS Sanity’s DOM engines</Text>
         </Box>
+        <Box marginY={5}>
+          <Button text="Open a hailing frequency" onClick={() => setDialogOpen(true)} />
+          {dialogOpen && (
+            <Dialog
+              id="ui3Dialog"
+              open={dialogOpen}
```
laurenashpole: Does ui3 expect an `open` prop? Then I'd lean even more towards `open` instead of `isOpen` for the new one.

packages/ui/src/components/modal/modal.css
```diff
@@ -0,0 +1,52 @@
+.sui-Modal {
+  flex-direction: column;
+  gap: var(--space-4);
+  overflow-y: auto;
```
laurenashpole: Since we're already using classes in `Modal.tsx`, could probably remove these three and add `sui-flex-direction-column sui-gap4 sui-overflow-y-auto`.

packages/ui/src/components/modal/modal.props.ts
```diff
@@ -0,0 +1,35 @@
+import {type PropDef} from '../../types/PropDef'
+import {type BoxProps} from '../box/box.props'
+
+/** @beta */
+export interface ModalProps extends Omit<React.ComponentProps<'dialog'>, 'open'> {
+  /** Text to be displayed as the modal's heading; optional but recommended for optimal accessibility and presentation */
+  header?: string
```
laurenashpole, line 6: I know you're still waiting on answers about the header from PJ but, if we're doing `Modal.Content` and `Modal.Footer`, it might make sense to switch this to `Modal.Header` once things are settled.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }, [isOpen])
+
+  return (
+    <dialog
```
laurenashpole, line 62: If we want to add a `size` prop to match `Container`, might be able to use that component here with `as="dialog"`.

packages/ui/src/components/modal/modal.props.ts
```diff
@@ -0,0 +1,35 @@
...
+    type: 'boolean',
+  },
+}
+
+/**
+ * Main content of the modal. Grows to fill the available space and scrolls if needed.
+ *
+ * @beta
+ */
+export type ModalContentProps<T extends React.ElementType = 'div'> = BoxProps<T>
```
laurenashpole: I'm not sure if we'd want to allow all `BoxProps` for `Modal.Content` or `Modal.Footer`.

apps/depth-testing/src/routes/ui3.tsx
colepeters: Fair, I'll rename :)

packages/ui/src/components/modal/modal.css
colepeters: Good call! I moved these from the `[open]` selector and totally missed being able to do this :D

packages/ui/src/components/modal/modal.props.ts
colepeters: Yeah let's definitely keep that in mind! I think that would be a good move

packages/ui/src/components/modal/Modal.tsx
colepeters: Yeah, let's dig into that once PJ’s back? I'm still really uncertain about that one

packages/ui/src/components/modal/modal.props.ts
colepeters: Oh that's a good spot — yeah I think probably just no props honestly 😄

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,139 @@
...
+import {Heading} from '../heading/Heading'
+import {IconButton} from '../icon-button/IconButton'
+import {type ModalProps, modalProps} from './modal.props'
+
+const modalClassName = suffixClassName('sui-Modal')
+const modalContentClassName = suffixClassName('sui-ModalContent')
+const modalFooterClassName = suffixClassName('sui-ModalFooter')
+
+function ModalRoot(props: ModalProps) {
+  // `open` must stay out of `rest`. Spread onto the dialog it would set the native open
```
laurenashpole: I don't think we need so much documentation around the `open` prop. For me at least, the comments make it seem more complicated than it actually is when it's really treated the same as any other `boolean` prop.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,139 @@
...
+      // In this branch, the `open` prop has changed to falsy;
+      // thus we need to fire the dialog element's `close` method,
+      // which in turn will fire the `onClose` function provided by the consumer
+      // (see dialog's `onClose` assignment in the render function below).
+      if (dialogElement.open) {
+        dialogElement.close()
+      }
+    }
+
+    // Ensure the modal is closed (and thus its `open` attribute updated)
```
laurenashpole, line 54: This comment seems to be explaining what a cleanup function which shouldn't be necessary.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,139 @@
...
+          onClick={() => modalRef.current?.close()}
+        />
+      </Flex>
+      {children}
+    </dialog>
+  )
+}
+
+/** Main content of the modal. Grows to fill the available space and scrolls if needed. */
+function ModalContent({children}: {children?: ReactNode}) {
```
laurenashpole: I think the types for these should probably be something like `ComponentProps<'div'>` since they should also accept all valid React props. And then they'd also need to use `getProps` for `className` and `style`. The `ListItemImage` subcomponent in [List.tsx](https://github.com/sanity-io/ui/blob/v5-alpha/packages/ui/src/components/list/List.tsx) is an example of a component with a specific element tag and no other props.

packages/ui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,139 @@
...
+    </Box>
+  )
+}
+
+ModalRoot.displayName = 'Modal'
+
+/**
+ * @beta
+ *
+ * Renders a modal dialog element — should not be used for nonmodal dialogs.
```
laurenashpole, line 130: I'm wondering if in general there's over commenting in this file? I think this sort of description should live in the documentation rather than a component's comments. The `Modal.Content` and `Modal.Footer` descriptions are mostly explained in their names (we repeat the names sometimes in type definitions because they're used in places like Storybook but that wouldn't be the case here). Just a thought, not blocking.

packages/ui/src/components/modal/Modal.tsx
colepeters: Fair, thanks! Will clean up

packages/ui/src/components/modal/Modal.tsx
colepeters: Makes sense, thanks!

packages/ui/src/components/modal/Modal.tsx
colepeters: Makes sense, yeah. There are some aspects (like this not being used for a nonmodal dialog) that I think might be helpful to have in editor overlays, but agreed a lot of this can be moved to proper docs. I'll clean up!

laurenashpole (approved):
LGTM! Thanks for going through all those type changes.

PR #2763, opened 2026-08-24
feat: v5 alpha popover and tooltip

laurenashpole:
### Description

Copied from https://github.com/sanity-labs/ui-poc/pull/123:

This PR makes some updates to the Tooltip component and introduces the Popover component. It also adds both to the INP testing app. 

A few important things to note:

- Both of these components use `cloneElement` to copy the trigger element. This was the cleanest solution for combining tooltips and popovers on the same trigger. I did a lot of performance testing compared to compound components and the cloned version always either won or was almost exactly the same. The results can be viewed in [this doc](https://docs.google.com/document/d/1849WKxGcRUZtlppas3kQfkdUrnrUpbyAYr-vl8liEcU/edit?tab=t.clne38y5kck#heading=h.it7ooktec54).
- After looking into the Popover API more, it made sense to update the Tooltip component to use that API with `interestfor` to open on hover. The upside is that this fixes the performance issues we were seeing with the previous version and now behaviors like close on escape are handled natively. One downside is that `interestfor` isn't fully supported across browsers so this does include a polyfill for it.
- The API for Tooltip changes a bit here. To match Popover, `text` becomes `content`. `anchorName` is needed to combine tooltips and popovers and I removed `disabled` because it can cause confusion there (and can be handled with conditional rendering).
- As part of the combining tooltips and popovers functionality, I added a flag to the components called `forwardsTriggerProps` and a utility to merge trigger props. These make sure that if a trigger is nested in a tooltip and popover (or vice versa), the trigger props (including style and event handlers) are forwarded and merged so both work.
- After starting on Menu, it became clear that these components might need to be rendered in portals so I added that optional prop and functionality. This includes a helper to render in a portal and one to detect client vs. server rendering.

### What to review

Both the Tooltip and Popover code should be reviewed.

### Testing

I added Storybook tests along with including these components in the INP and mount testing. Once this is published as a beta, I want to try it the components out in Studio to double check that they work in more complicated situations and that the placements match.

---

> [!NOTE]
> **Medium Risk**
> Breaking Tooltip API changes and reliance on Popover/`interestfor` (with polyfill) affect overlay behavior across apps; nested trigger merging is easy to get wrong in edge cases.
> 
> **Overview**
> Introduces a **beta `Popover`** built on the native Popover API (`popover="auto"`), with optional **`portal`** rendering, **`anchorName`** for shared anchors, and **`content`** as the body prop. **`Tooltip`** is reworked to the same model: **`text` → `content`**, **`disabled` removed**, **`popover="hint"`** plus **`interestfor`** for hover (with an optional **`@sanity/ui/polyfills`** entry loading the **`interestfor`** polyfill).
> 
> Both components use **`cloneElement`** and a **`forwardsTriggerProps` / `mergeTriggerProps`** pattern so tooltip and popover can wrap the same trigger. Shared helpers **`renderPortal`** and **`useIsClient`** support SSR-safe portals. Styling shifts to **`:popover-open`** / **`@starting-style`** and shared placement rules; **`TooltipGroup`** now keys off opacity transitions and sets **`--tooltip-delay-group`** to **`0ms`**.
> 
> Storybook gains **Popover** stories and updated tooltip tests; **inp-testing** and **mount-testing** add bulk mount benchmarks and migrate to **`content`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e341a5e338005874e91fedd8efaa7fc620cf13a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ui/src/hooks/useIsClient.ts
```diff
@@ -0,0 +1,10 @@
+import {useSyncExternalStore} from 'react'
+
+/** @public */
+export function useIsClient() {
+  return useSyncExternalStore(
+    () => () => {},
+    () => true,
+    () => false,
+  )
+}
```
stipsan, line 10: A future React version is finally adding a first-class API for this: https://react.dev/reference/react-dom/browser

stipsan (approved):
Stellar and comprehensive work, really! 💖

packages/ui/src/hooks/useIsClient.ts
laurenashpole: Lol, I don't know why Cursor decided to get involved here.

PR #2800, opened 2026-08-27
fix: support cross-file transformations in all codemods

laurenashpole:
### Description

This PR updates the all the remaining codemods to support cross-file styled component transformations. Again, there's some repetition, especially in the tests, but for now I'm not sure if it's worth a major refactor.

### What to review

Make sure the logic across the files seems consistent.

### Testing

I updated the tests for each codemod to include the cross-file transformations.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to @sanity/ui-codemod test and transform logic with no runtime app impact; incorrect alias resolution could mis-migrate consumer code during upgrades.
> 
> **Overview**
> Extends **cross-file styled-component migration** to the remaining typography/layout codemods (**Card**, **Code**, **Heading**, **Label**, **Text**), matching the approach already used for **Stack**.
> 
> Each transform now resolves **`styled(Component)` aliases** via `getStyledComponentAliases`, treats those usages in `shouldTransformComponent`, and runs **`transformStyledComponents`** so JSX on imported wrappers (including **barrel re-exports**) gets the same attribute mods, TODO comments, and **trim/as** additions as direct component usage. **Card** and **Label** also pass **`localNames`** into `replaceElement` and share a **`replaceWithBox`** helper so styled wrappers follow the same Card→Box rules; **Stack** only renames **`hasFlexAttrs`** to **`replaceWithFlex`** for consistency.
> 
> **`defineCrossFileTest`** coverage was added per codemod, including cases that **skip homonymous components from other packages**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b5d3753792e7c2e816ca90dd22d5053463797da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ui-codemod/src/transforms/latest/card/card.test.ts
+++ b/packages/ui-codemod/src/transforms/latest/card/card.test.ts
@@ -1,4 +1,6 @@
-import {defineInlineTest} from '../../../utils/testUtils'
+import {expect} from 'vitest'
+
+import {defineCrossFileTest, defineInlineTest} from '../../../utils/testUtils'
 import transform from './card'
 
 defineInlineTest(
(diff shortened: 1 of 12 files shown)


packages/ui-codemod/src/transforms/latest/card/card.test.ts
```diff
@@ -205,3 +207,100 @@ defineInlineTest(
...
+  `
+    import {Card} from '@sanity/ui'
+
+    export const RootCard = styled(Card)(({theme}) => ({}))
+  `,
+  `
+    import {RootCard} from './Component.styled'
+
+    export function Component() {
+      return <RootCard padding={4} radius={3} border />
```
colepeters, line 223: Just curious: do you think we should warn if `border` is set to false or omitted?

packages/ui-codemod/src/transforms/latest/text/text.test.ts
```diff
@@ -111,3 +113,84 @@ defineInlineTest(
...
+  `
+    import {RootText} from './Component.styled'
+
+    export function RootCode() {
+      return <RootText flex="auto" />
+    }
+  `,
+  (output) => {
+    expect(output.replace(/\s+/g, ' ')).toContain(
+      '<RootText style={{ flex: "1 1 auto" }} as="div" trim={true} />',
```
colepeters, line 134: Just wondering, do we need to test for any `as` prop changes carrying over, or should testing for one transformed prop cover any prop changes/carryovers? (Thinking about how often `as` might be used with `Text`)

colepeters (approved):
LGTM! A couple non blocking questions, feel free to merge!

packages/ui-codemod/src/transforms/latest/card/card.test.ts
laurenashpole: Right now a `border` is required to qualify as a Card component. If it's missing, the component will be replaced with a Box or, for the cross-file styled components, set a `Please double check styled(Card) migration below` warning. That's covered in the test after this one.

packages/ui-codemod/src/transforms/latest/text/text.test.ts
laurenashpole: Testing for a user setting an `as` prop is already included in the existing transformation tests. It should be covered by `does not add as prop if already set`.

packages/ui-codemod/src/transforms/latest/card/card.test.ts
colepeters: Ahh perfect, thanks!

packages/ui-codemod/src/transforms/latest/text/text.test.ts
colepeters: Sweet! Thanks!

PR #2878, opened 2026-09-08
feat: add topAnchor prop for Dialog

colepeters:
### Description

@georgedoescode recently requested the ability to anchor Dialogs with an offset from the top of the viewport rather than centre them, given several instances in Media Library where this is currently done.

This PR implements that feature via the optional `topAnchor` prop. If the prop is not provided, Dialogs remain centred on the viewport. When provided, the Dialog will be positioned the provided length from the top of the viewport, remaining centred on the inline axis. We also set a complementary max height on the Dialog, so that it cannot grow taller than `100dvh - (topAnchor * 2)`) which effectively means a Dialog maxed out in height would have the same offset from the bottom of the viewport as the top.

### What to review

- New topAnchor prop + CSS, plus propDef on Dialog
- New + updated stories for coverage

### Testing

- Verified in Storybook and added Storybook tests to ensure the class is added when the prop is provided, and that it's not when the prop is undeclared

---

> [!NOTE]
> **Low Risk**
> Optional layout prop and CSS with no changes to open/close or modal behavior when `topAnchor` is omitted.
> 
> **Overview**
> Adds an optional **`topAnchor`** prop on `Dialog` so consumers can pin the modal a fixed distance from the top of the viewport instead of vertically centering it. When `topAnchor` is set, horizontal centering is preserved (`sui-mx-auto`) and a new **`sui-top-anchor`** class applies `inset-block-start` via `--top-anchor`, plus a **`max-block-size`** so tall dialogs keep symmetric spacing from the bottom.
> 
> Default behavior is unchanged: without the prop, dialogs stay centered as before. Storybook gains a **With top anchor** story and play tests assert **`sui-top-anchor`** is present only when the prop is used.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f12d8b88a945135cc412af0fd8fe6aa971a17f8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ui/src/css/classes/local/index.css
+++ b/packages/ui/src/css/classes/local/index.css
@@ -3,3 +3,4 @@
 @import './text-overflow.css';
 @import './text-trim.css';
 @import './tone.css';
+@import './top-anchor.css';
--- a/packages/ui/src/css/classes/local/top-anchor.css
+++ b/packages/ui/src/css/classes/local/top-anchor.css
@@ -1,4 +1,4 @@
 .sui-top-anchor {
   inset-block-start: var(--top-anchor);
-  max-height: calc(100dvh - var(--top-anchor) * 2);
+  max-block-size: calc(100dvh - var(--top-anchor) * 2);
 }
--- a/packages/ui/src/css/classes/local/top-anchor.css
+++ b/packages/ui/src/css/classes/local/top-anchor.css
@@ -1,0 +1,4 @@
+.sui-top-anchor {
+  inset-block-start: var(--top-anchor);
+  max-height: calc(100dvh - var(--top-anchor) * 2);
+}
--- a/.changeset/spotty-bananas-work.md
+++ b/.changeset/spotty-bananas-work.md
@@ -1,0 +1,5 @@
+---
+'@sanity/ui': patch
+---
+
+add topAnchor prop for Dialog
--- a/apps/storybook/src/stories/Dialog.stories.tsx
+++ b/apps/storybook/src/stories/Dialog.stories.tsx
@@ -122,12 +122,33 @@ function DialogWithFooterStory(props: DialogProps) {
   )
 }
 
+function DialogTopAnchorStory(props: DialogProps) {
+  const [dialogOpen, setDialogOpen] = useState(false)
+
+  return (
+    <Box padding={4}>
+      <Button text="Open dialog" onClick={() => setDialogOpen(true)} />
+      <Dialog
+        {...props}
+        open={dialogOpen}
+        onClose={() => setDialogOpen(false)}
+        topAnchor="clamp(1rem, 10dvh, 4rem)"
+      >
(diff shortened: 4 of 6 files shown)


packages/ui/src/css/classes/local/top-anchor.css
```diff
@@ -0,0 +1,4 @@
+.sui-top-anchor {
+  inset-block-start: var(--top-anchor);
```
laurenashpole, line 2: Just curiosity since they seem the same visually, is there a reason to use `inset-block-start` here instead of `top`? I tried looking it up and it seemed like the main difference was different layouts for different languages or text directions?

packages/ui/src/css/classes/local/index.css
```diff
@@ -3,3 +3,4 @@
 @import './text-overflow.css';
 @import './text-trim.css';
 @import './tone.css';
+@import './top-anchor.css';
```
laurenashpole, line 6: Wondering if we'll ever use this outside of dialog? Not really a big deal though.

laurenashpole (approved):
Looks good to me! Left a couple questions but otherwise approved!

packages/ui/src/css/classes/local/top-anchor.css
colepeters: No functional difference for this unless for some reason this is rendered in an environment where writing direction is set to `sideways-lr`, which none of our apps would support, and which would be extremely rare outside of that.

So really it just comes down to: I try and make use of logical properties wherever possible, as eventually I think we should work on supporting non-LTR languages, and consistently using logical properties is helpful for that. ☺️

packages/ui/src/css/classes/local/index.css
colepeters: I'm guessing not! I just noticed that pretty much all of our prop defs that generate classes are defined in `classes/local`. I'd be fine w/ moving into the Dialog CSS though — not a strong opinion!

packages/ui/src/css/classes/local/top-anchor.css
laurenashpole: That makes sense. Does it matter if the specific element should reposition based on writing direction (meaning, should the dialog layout in particular change for `sideways-lr` or would we want it in the same spot even then)?

colepeters: I'd need to do more research to answer this confidently. My current understanding is basically, in top to bottom languages, we expect that information surfaced at the top of a document to be information that is important, because it comes at the start of a document's flow.

So my presumption is in a document where the writing mode is 'sideways' and thus the block direction is physically left to right, 'importance' would be denoted by something being at the start — so, physically the left — of the document.

But really for me this is more about the consistency of using logical directions 😄

PR #2884, opened 2026-09-09
refactor(ui): track LayerProvider child layers with a single useReducer

stipsan:
## Summary

`LayerProvider` kept two `useState` values, `childLayers` (a per-level count) and `size` (the number of distinct child levels), and synced them by calling `setSize(Object.keys(nextState).length)` from inside the `setChildLayers` updater. This replaces both with one pure reducer in `layerReducer.ts`.

The state has two fields that each count one thing. `childLayers` is a `ReadonlyMap` from a level to the number of registered child layers on it. `childrenWithoutLevel` counts children that call `registerChild()` without a level, which older `@sanity/ui` copies sharing the globally scoped `LayerContext` do. `LayerProvider` derives `size` as `childLayers.size + childrenWithoutLevel`, so the reducer has no invariant to keep in sync and the derivation allocates nothing.

Registering or unregistering a child is now a single state update instead of two. `registerChild` depends only on `parentRegisterChild`, because `dispatch` is stable, so its identity does not change as child layers mount and unmount.

The reducer's `switch` checks exhaustiveness with `action satisfies never` in its `default` branch and returns `state` there, so a new action variant fails to compile until handled.

## Tests

`layerReducer.test.ts` covers the state transitions: adding a level, counting further children on the same and on different levels, counting children without a level, dropping a level when its last child unregisters, returning to the initial state, and not mutating the previous state.

`layerProvider.test.tsx` renders `LayerProvider` and `Layer` trees under `StrictMode` and asserts `level`, `size`, `isTopLayer` and `zIndex` per layer: a root with no children, nested levels with accumulated `zIndex`, siblings counting as one level, `isTopLayer` restored as children unmount, children that register without a level, and `registerChild` identity staying stable.

## Verification

`pnpm --filter @sanity/ui test` passes (124 tests), as do `pnpm lint`, `pnpm knip` and `oxfmt --check`. The Storybook browser tests for the `Layer`, `Dialog`, `Menu`, `MenuButton`, `Popover`, `Toast`, `Tooltip` and `Autocomplete` stories pass in headless Chromium, including the `Layer` `Nested` play test that asserts `size=1` then `size=2` as nested layers open.

Includes a patch changeset for `@sanity/ui`.

<div><a href="https://cursor.com/agents/bc-90672e4f-0fcb-4fc7-aa95-9d2ac454836a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-90672e4f-0fcb-4fc7-aa95-9d2ac454836a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/ui/src/core/utils/layer/layerReducer.ts
+++ b/packages/ui/src/core/utils/layer/layerReducer.ts
@@ -1,0 +1,77 @@
+/**
+ * The layers registered below a `LayerProvider`.
+ *
+ * @internal
+ */
+export interface LayerState {
+  /** Number of registered child layers per level (`level -> count`). */
+  childLayers: Record<number, number>
+  /**
+   * Number of levels that have at least one registered child layer, plus one per child that
+   * registered without a level. A size of `0` means the layer is the top layer.
+   */
+  size: number
+}
+
+/**
+ * Registers or unregisters a child layer. Children from older `@sanity/ui` copies that share the
+ * layer context register without a `level`.
+ *
+ * @internal
+ */
+export type LayerAction =
+  | {type: 'child/register'; level?: number}
+  | {type: 'child/unregister'; level?: number}
+
+/**
+ * @internal
+ */
+export const initialLayerState: LayerState = {childLayers: {}, size: 0}
+
+/**
+ * @internal
+ */
+export function layerReducer(state: LayerState, action: LayerAction): LayerState {
+  const {level} = action
+
+  switch (action.type) {
+    case 'child/register': {
+      if (level === undefined) {
+        return {...state, size: state.size + 1}
+      }
+
+      const count = state.childLayers[level] ?? 0
+
+      return {
+        childLayers: {...state.childLayers, [level]: count + 1},
+        size: count === 0 ? state.size + 1 : state.size,
+      }
+    }
+
+    case 'child/unregister': {
+      if (level === undefined) {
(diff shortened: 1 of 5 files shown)


packages/ui/src/core/utils/layer/layerReducer.ts
```diff
@@ -0,0 +1,77 @@
...
+  switch (action.type) {
+    case 'child/register': {
+      if (level === undefined) {
+        return {...state, size: state.size + 1}
+      }
+
+      const count = state.childLayers[level] ?? 0
+
+      return {
+        childLayers: {...state.childLayers, [level]: count + 1},
```
stipsan: could it be more performant if `childLayers` were a `Map`?

PR #2887, opened 2026-09-09
Migrate the UI build to tsdown and native React Compiler

stipsan:
## Summary

- migrate only `packages/ui` from `@sanity/pkg-utils` to tsdown and `@sanity/tsdown-config`
- compile the UI package, Vite apps, and Next example with the native Rust React Compiler
- upgrade oxfmt, oxlint, Vite, and their related Oxc tooling
- preserve `packages/ui-codemod`, the UI package exports, CSS subpath, and CLI build
- declare `'use client'` in the two entry sources instead of injecting a build banner
- suppress the two new Storybook addon findings inline rather than reworking the components

## Build timings

Median wall time on the same Node 24.14 VM after one discarded warm-up, with Turbo cache bypassed for root builds:

| Build | Before | After | Delta |
| --- | ---: | ---: | ---: |
| `packages/ui` JavaScript, 5 runs | 3.821s | 2.266s | -41% |
| Root packages, 3 runs | 11.463s | 10.155s | -11% |
| Storybook, 3 runs | 6.642s | 5.051s | -24% |

## Output comparison

- identical 35 runtime exports and 106 type exports
- identical external dependency set and 40 React Compiler cache calls
- `dist/index.js` reduced from 75,219 bytes to 70,639 bytes, about 6%
- both published entries keep the leading `"use client"` directive, now hoisted from source
- the lazy refractor chunk becomes content-hashed and no longer repeats the directive, since it is only reachable through a dynamic import inside the client entry and is not exposed by the `exports` map

## Verification

- `pnpm lint`
- `pnpm ts:check`
- `pnpm build --force`
- `pnpm test`
- tracked-file `oxfmt --check`
- UI 85/85 tests, ui-codemod 209/209 tests, Storybook Chromium 90/90 tests
- Vite, React Router, Storybook, and Next native-compiler builds
- runtime, type export, external, compiler-cache, and client-directive parity

<div><a href="https://cursor.com/agents/bc-3a2ae2e8-81f9-4f57-a2ff-6484f0b1c3c6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3a2ae2e8-81f9-4f57-a2ff-6484f0b1c3c6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/storybook/src/addons/performance/withProfiler.tsx
```diff
@@ -17,6 +17,7 @@ const channel = addons.getChannel()
 
 const HIDDEN_BENCHMARK_STYLE: CSSProperties = {
   border: 0,
+  // oxlint-disable-next-line typescript/no-deprecated
   clip: 'rect(0, 0, 0, 0)',
```
laurenashpole:
```suggestion
clipPath: 'rect(0, 0, 0, 0)',
```

laurenashpole (approved):
LGTM!

PR #2930, opened 2026-09-18
feat(v5): add Select component

colepeters:
### Description

Adds the new v5 Select component.

Removed props:

- `fontSize` — now renders at the fixed `body1` text size.
- `padding` — replaced by `density` (`regular` or `loose`).
- `radius` — the select now uses a fixed radius.
- `gap` — the spacing between the control and the icon is now fixed.
- `space` — a deprecated alias for `gap`; removed.
- `customValidity` — no longer supported (TBD if this is okay w/ Studio team)

New props:

- `density` — `regular` (default) or `loose`
- `hasError` — toggles error state styling

Behaviour changes:

- `readOnly` no longer has an effect. The old component treated it as `disabled`. The `readonly` attribute is not valid on a select element.
- `hasError` sets `aria-invalid` on the select and adds the `sui-error` class. Improves screen-reader support for the invalid state.
- `disabled` and all other native select attributes still pass through.

### What to review

- Anything missing or odd in the codemod?
- All other feedback welcome :)

### Testing

- Accessibility tests pass in Storybook
- Codemod tests pass

---

> [!NOTE]
> **Medium Risk**
> Adds a public form control with a smaller prop surface and behavior changes (e.g. no `customValidity`, `readOnly` ineffective), so consumers rely on the codemod and manual follow-up for edge cases.
> 
> **Overview**
> Introduces a **v5 `Select`** in `@sanity/ui`: a styled native `<select>` with a fixed dropdown icon, **`density`** (`regular` / `loose`), **`hasError`** (`aria-invalid` + error styling), and passthrough for standard select attributes and children.
> 
> The legacy styling/validation API is narrowed—**`fontSize`**, **`padding`**, **`radius`**, **`gap`**, **`space`**, **`customValidity`**, and effective **`readOnly`** are dropped or changed—and a **`@sanity/ui-codemod` `latest/select`** transform updates imports, removes **`space`**, and emits TODO comments for props that need manual migration (including styled wrappers).
> 
> Also adds **Storybook** coverage (default, disabled, error, interaction/a11y plays), wires **CSS** into the package bundle, exports **`FormElementDensity`**, and sets **disabled form controls to `opacity: 1`** in the global reset so disabled selects match the new styles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8fb0145e3af10016418d8571eb78ff78c4602a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ui-codemod/src/transforms/latest/select/select.mods.ts
```diff
@@ -0,0 +1,36 @@
+import type {AttributeMods} from '../../../types/AttributeMods'
+
+/** @internal */
+export const SELECT_MODS: AttributeMods = {
+  customValidity: {
+    type: 'warn-only',
+    warning:
+      'Please double check the Select migration below. The customValidity prop is no longer supported. Handle validation externally and pass hasError to toggle the invalid styling.',
+  },
+  fontSize: {
```
laurenashpole, line 10: For this and the other styling props that are being deprecated, the pattern in other migrations would be to use `style-only` or `style-mapped` since they should be migratable without a user needing to intervene.

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
```diff
@@ -0,0 +1,36 @@
+import type {AttributeMods} from '../../../types/AttributeMods'
+
+/** @internal */
+export const SELECT_MODS: AttributeMods = {
+  customValidity: {
```
laurenashpole, line 5: Should this be migrated to `hasError`?

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
```diff
@@ -0,0 +1,36 @@
...
+    type: 'warn-only',
+    warning:
+      'Please double check the Select migration below. The padding prop is no longer supported. Use the density prop (regular or loose) instead.',
+  },
+  radius: {
+    type: 'warn-only',
+    warning:
+      'Please double check the Select migration below. The radius prop is no longer supported. Select uses a fixed radius.',
+  },
+  readOnly: {
```
laurenashpole, line 30: Should this be migrated to `disabled`?

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
```diff
@@ -0,0 +1,36 @@
...
+    type: 'warn-only',
+    warning:
+      'Please double check the Select migration below. The readOnly prop is no longer valid, as the readonly HTML attribute is invalid on select elements.',
+  },
+  space: {type: 'remove'},
```
laurenashpole, line 35: This should probably be a `style-mapped` migration with the same values as `gap`.

packages/ui-codemod/src/transforms/latest/select/select.test.ts
```diff
@@ -0,0 +1,164 @@
+import {expect} from 'vitest'
```
laurenashpole, line 1: I think the all or most of the migrations tested in this file should be migrations instead of warnings.

packages/ui/src/components/select/select.css
```diff
@@ -0,0 +1,67 @@
...
+  --border-color: var(--separator-high);
+  --color: var(--foreground-high);
+  background-color: var(--background-color);
+  border: 1px solid var(--border-color);
+  font: var(--form-element-text);
+  color: var(--color);
+}
+
+[data-ui="Select-picker-icon-slot"] {
+  --background-color: var(--backdrop);
```
laurenashpole, line 12: Is there a reason you need a background color for the icon? It seems like it should be fine being transparent and without the `calc`s for positioning.

packages/ui/src/components/select/Select.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {suffixClassName} from '../../utils/suffixClassName'
+import {Icon} from '../icon/Icon'
+import {type SelectProps, selectProps} from './select.props'
+
+const selectClassName = suffixClassName('sui-Select')
+
+/** @public */
+export function Select({
+  density = 'regular',
+  disabled = false,
```
laurenashpole, line 15: Just for simplicity, since `disabled` is false you shouldn't need to set a default for it. In that case, we never need to use it explicitly since it'd be in the `props` spread below.

packages/ui/src/components/select/Select.tsx
```diff
@@ -0,0 +1,57 @@
...
+        style={{...style, anchorName}}
+        {...rest}
+        aria-invalid={hasError || undefined}
+        disabled={disabled}
+      >
+        {children}
+      </select>
+      <div
+        className={iconClasses}
+        data-ui="Select-picker-icon-slot"
```
laurenashpole, line 48: I'm not sure if this really matters but are the uppercase in other components? Also, for consistency with the main component, could it be shortened to something like "Select-Icon"?

laurenashpole (changes requested):
Looking good!

I left a few comments on the codemod because I think we could still make migrations a little easier for users. Otherwise, just some minor questions/comments on the component itself.

packages/ui/src/components/select/select.css
colepeters: It's to prevent the current selected option from visually running under the icon. Without a background, the icon would sit over top of the text; with a background, it's blocked out:

image

colepeters: (This could also be solved with padding on the select element, but since padding is already a little complex w/ the density setup, I didn't want to make it even more complicated, esp. with it potentially being different at different densities.)

packages/ui/src/components/select/Select.tsx
colepeters: Good call, thanks!

packages/ui/src/components/select/Select.tsx
colepeters: We use uppercase for component names (e.g. `data-ui="Select"`) so I figured it'd be consistent here for a segment/child of a component. Not a tightly held opinion though!

I'm down with `Select-icon` but wasn't sure if in the future we might add an Icon slot/prop to Select. But probably no need to complicate it now if that isn't currently a factor. I'll shorten it!

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
colepeters: `space` is deprecated in UI3/4 so I think this one should be okay to remove?

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
colepeters: I'm not sure on this one — I think that's a reasonable call, but I'm wondering if it might need more work on the consumer side since disabling a select will remove it from the accessibility tree, and it might need to remain in there. That said, readOnly selects are currently disabled, so it's not making anything worse. WDYT?

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
colepeters: It's possible, but I'm not sure if it's always a clear 1:1. Consumers might think we'll handle the error message for them if we migrate it, and further if they're passing a string (which is what the prop is currently for) it will get evaluated as a boolean in the component, which might be weird. I'm leaning towards warning on this one so consumers make the call. WDYT?

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
colepeters: Thanks! I've switched several of these over to inline style migrations, but some (like `gap`) don't have any effect on the new component so likely would be weird to carry over.

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
laurenashpole: It's marked as deprecated in v3 but that happened pretty recently and it still works so users might still be using it. Since it maps to gap, we could include it to make v3 to v5 migrations easier.

packages/ui/src/components/select/Select.tsx
laurenashpole: For some reason I was thinking all words were capitalized (also just realized I wrote uppercase when I meant capitalized) but, I don't know if there are examples or if I imagined that and it doesn't really matter much either way. I think for the postcss plugin it only matters if the first letter is capitalized.

packages/ui/src/components/select/select.css
laurenashpole: That makes sense, I wasn't thinking about longer options.

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
laurenashpole: Wait, I just read your comment on another thread. If gap doesn't do anything, may as well remove it. In later components like TextInput we might want to include both.

colepeters: Oh yup, good call!

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
laurenashpole: That's probably fine if it's not necessarily 1:1.

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
laurenashpole: If we're currently using `readOnly` to set the `disabled` attribute it could be updated since the behavior is the same. I'm wondering if we would ever want to support both?

laurenashpole (approved):
LGTM! Thanks for making those codemod updates!

packages/ui-codemod/src/transforms/latest/select/select.mods.ts
colepeters: We currently set `disabled` to prevent editing, but `readOnly` isn't supported by the select element, so we wouldn't ever want to support both — the main thing, I think, is whether consumers might need to display the value of the select in a different way, to allow screen readers to capture it without it being editable.

